PR #5, opened 2015-03-05
added a package level logger in log.py and imports into package files

yosemitesammy217:
added a package level logger in log.py and imports into package files

the logger is not yet used within any of the asyncssh classes or functions, but this puts in place the structure to emit messages to a logger named "asyncssh" where the user of asyncssh may set logging levels and format the messages as they wish.

usage within a function of asyncssh:

logger.debug("some debug message")
logger.info("some info message")

to use in a program:

assh_logger = logging.getLogger("asyncssh")

assh_logger.setLevel(logging.DEBUG)


ronf:
This is now merged into my develop branch, and I'll move it to master shortly.

PR #7, opened 2015-03-18
Impove handling of known_hosts file.

alexander-travov:
Skip comment lines in known_hosts.
Add support for markers (@cert-authority, @revoked).
Add support for negated patterns.
Add support for hashed form.

Links that may be of interest:

- sshd man page:
  http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man8/sshd.8?query=sshd
  Section SSH_KNOWN_HOSTS FILE FORMAT

- openssh implementation of hashed entries:
  https://github.com/openssh/openssh-portable/blob/master/hostfile.c

- Twisted conch implementation:
  https://twistedmatrix.com/trac/browser/trunk/twisted/conch/client/knownhosts.py

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -23,6 +23,7 @@ from .cipher import *
 from .compression import *
 from .forward import *
 from .kex import *
+from .known_hosts import parse_known_hosts
 from .listen import *
 from .logging import *
 from .mac import *
@@ -1418,7 +1419,7 @@ class SSHClientConnection(SSHConnection):
         else:
             if server_host_keys is () and server_ca_keys is ():
                 server_host_keys, server_ca_keys = \
-                    self._parse_known_hosts(host, port)
+                    parse_known_hosts(host, port)
             else:
                 server_host_keys = self._load_public_key_list(server_host_keys)
                 server_ca_keys = self._load_public_key_list(server_ca_keys)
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for submitting this, alexander! If I accept this code, are you ok if I release it under the Eclipse Public License (matching the license on the other files here)?

alexander-travov:
Yes

ronf:
An update on this: I am planning to include all the capabilities here, but I've found a couple of issues in the way the pattern matching is done here that I'm working on, to try and better match OpenSSH's behavior.

In particular, OpenSSH allows wildcards even on entries with non-standard ports, and the wildcard can match a mix of characters from the host portion & the port portion. The Python fnmatch() function needs square brackets to be escaped in patterns to match them as literals, though. Otherwise, they are treated as character classes.

OpenSSH also seems to do a lookup without the port number in some cases if no matching entry was found with the port number. So, if you have an entry for "localhost" but you are connecting on port 2222 it will look for [localhost]:2222 first but then try looking for plain localhost after that if it doesn't find a match. I'm still trying to figure out how this interacts with negated host patterns, though.

If we add support for @revoked here, I would also like to do something more than just ignore those keys, ideally returning them to the calling code so AsyncSSH could fail with something like a 'RevokedKeyError' exception if a server tries to present a revoked key.

Once I have these issues worked out, I'll be checking in an update that should cover the functionality in this pull request, but the implementation of large parts of it will be new.

alexander-travov:
I made first implementation based only on my interpretation of sshd man page and sporadic tests.
Now added support for glob patterns usage in entries with nonstandart port.

Looking into creating tests for known_hosts functionality:
- [Paramiko known_hosts tests](https://github.com/paramiko/paramiko/blob/master/tests/test_hostkeys.py)
- [Twisted conch known_hosts tests](https://twistedmatrix.com/trac/browser/trunk/twisted/conch/test/test_knownhosts.py)

alexander-travov:
+1 on explicit revoked keys handling

ronf:
@alexander-travov: I've just checked in a first cut of this. It includes revoked key handling, and some improvements to pattern matching on non-standard ports. It also handles the fallback to look for entries without a port number when no keys match, without having to parse the file twice.

I haven't had a chance to look through the known_hosts tests above yet, so some additional tweaks may be needed. However, I wanted to get this checked into my develop branch so you could look it over, Alexander. Thanks very much for your contribution, and please let me know if you see any problems in the code!

PR #8, opened 2015-03-18
Exclude usual suspects

alexander-travov:


ronf:
I have a local .gitignore with references to local files I don't want to ever commit, so I'd prefer not to have a .gitignore in the public repo.

alexander-travov:
You can use .git/info/exclude to ignore files in your local copy.
https://help.github.com/articles/ignoring-files/#explicit-repository-excludes

alexander-travov:
[python gitignore template](https://github.com/github/gitignore/blob/master/Python.gitignore)

nchammas:
FWIW @ronf, a .gitignore in the public repo is useful for ignoring files that all users or developers are likely to generate in the course of interacting with the project. That way people don't have to figure out the right .gitignore to have over and over by themselves.

Most projects I know include their .gitignore in the repo to cover such things. Here, for example, are [Apache Spark's .gitignore](https://github.com/apache/spark/blob/master/.gitignore) and [Fabric's .gitignore](https://github.com/fabric/fabric/blob/master/.gitignore).

ronf:
Ah, thank you both for the info on this. I'm still somewhat new to git and wasn't familiar with the explicit repository excludes. That's perfect for what I needed here. I've gone ahead and checked in a .gitignore that contains what most folks should need in both the master and develop branches!

Issue #18, opened 2015-05-31
AsyncSSH's dependencies are not visible to pip


nchammas:
pip thinks that AsynsSSH does not have any dependencies.

```
$ pip3 show asyncssh
---
Metadata-Version: 1.1
Name: asyncssh
Version: 1.1.1
Summary: AsyncSSH: Asynchronous SSHv2 client and server library
Home-page: http://asyncssh.timeheart.net
Author: Ron Frederick
Author-email: ronf@timeheart.net
License: Eclipse Public License v1.0
Location: /usr/local/lib/python3.4/site-packages
Requires: 
```

This may well be the case, but there are several modules [mentioned in the README](https://github.com/ronf/asyncssh/blob/8ea2dd5ead4a027685c4bc157370a8955d9ca9a5/README.rst#installation) that the user has to install to enable additional functionality.

If you expect a large proportion of the user base to need any of these modules, I suggest capturing them as dependencies in PyPI so that pip will automatically install them when it installs AsyncSSH.

This greatly simplifies things for your users. I would venture that most people would expect things to just work after a single `pip install`.

For comparison, this is what pip sees for Pandas (note the `Requires:` line):

```
$ pip3 show pandas
---
Metadata-Version: 2.0
Name: pandas
Version: 0.16.0
Summary: Powerful data structures for data analysis, time series,and statistics
Home-page: http://pandas.pydata.org
Author: The PyData Development Team
Author-email: pydata@googlegroups.com
License: BSD
Location: /usr/local/lib/python3.4/site-packages
Requires: python-dateutil, pytz, numpy
```

nchammas:
Specifically, I think this requirement at the least

> Install PyCrypto 2.6 or later from http://www.pycrypto.org and/or
   PyCA 0.6.1 or later from https://cryptography.io to provide basic
   crypto support.

should be captured programmatically and struck from the README. (In fact, I'm currently working on a pull request with several proposed refinements to the README. This is one of them.)

And I, for one, wouldn't mind the optional requirements like libsodium and bcrypt also be captured programmatically, given they don't drastically affect the install time of AsyncSSH.

ronf:
I have actually captured all of the dependencies (pycrypto, pyca, bcrypt, and libnacl) already in setup.py. However, I haven't made any of them required dependencies just yet. For now, I have them listed as "extras", and let you decide which ones you want to install. It's easy to choose them if you want, though. You'd just do something like:

[...]

PR #19, opened 2015-06-01
Refine README

nchammas:
This PR improves the README a bit by taking cues from the READMEs of [some extremely popular projects](https://github.com/kennethreitz/requests#requests-http-for-humans).

Specifically:
* The README now opens with a quick AsyncSSH example. The idea is to give prospective users something extremely minimal that they can immediately see, judge, and even copy into the REPL to experiment with.
* The License section, not really a thing of immediate interest to most users, is now at the end of the README.
* Section titles are now formatted accordingly. (Previously, they were formatted at the same level as the main title.)
* The installation instructions are simplified, partly in expectation of a resolution to #18.
* Finally, I removed an apparently duplicate README file and updated the examples link.

In addition to these proposed changes, I would suggest cutting down the Features section to roughly a third of its current size, and ordering the list by importance to users. This is important since, past the opening example, the feature list is how users know whether the library is right for them. And they're generally going to skim quickly for some key feature or two.

Cutting down the feature list to the most important subset can be very tough because a) the amount of effort put into implementing a feature is often not commensurate with how much users appreciate it; and b) your users are often very different from you as the author.

Since I don't have a good sense of the intended AsyncSSH user base, I didn't attempt this. But for starters, I would give more prominence to "big stroke" features like SFTP support and relegate details like what ciphers are supported to a minor subsection, or move them to a separate page entirely (like in the docs).

Side note: I am opening this PR against the `develop` branch, since I believe that is where changes are supposed to start. If so, I would add a note about this to contributing guide. (Actually, on that note, why don't changes just land directly in master? The projects I am familiar with, like [Apache Spark](https://github.com/apache/spark) for example, do this.)


ronf:
Thanks - I'll go through this soon and do the cleanups you have suggested here.

I like the idea of opening with a simple example.

Regarding the license, I'm not sure about moving that all the way to the bottom of the page. I kind of want people to think about how the package in licensed before they decide to use it, but I'll see if I can find a good place for this that's a bit further down but not too buried.

In terms of the installation instructions, let me know what you think in the other thread about my use of setuptools extras as a way to deal with this, at least for now.

On the duplicate README, you're right -- I should have removed that when I switched over to the README.rst. I added the original README before I was doing any of the Sphinx-based documentation.

On the feature list, I'll see what I can do to trim that back, perhaps moving some of the details down into the main documentation as you suggested. It started off as a pretty small list, but it has grown quite a bit in the last few releases.

Also, thanks for catching the out-of-date examples link and the section formatting issue.

Regarding your question about the branches, my intent it to keep "develop" for things that I want to make available to developers to try out before I'm ready to actually put them in an official release. One of the reasons for this is to experiment with the API, with no guarantee that anything added in "develop" won't change before it is released, and no guarantee of backward compatibility. Once I move things over to "master" and cut a new release, I would try to maintain backward compatibility (perhaps eventually with some deprecation mechanism).

Based on this, most new features will probably be added to "develop" first, but some bug fixes or simple improvements may get dropped directly into "master" first and then be merged forward into "develop".

nchammas:
> Regarding the license, I'm not sure about moving that all the way to the bottom of the page. I kind of want people to think about how the package in licensed before they decide to use it, but I'll see if I can find a good place for this that's a bit further down but not too buried.

That's fair. Actually, it's good that you want people to think about it.

Perhaps then it would be good to summarize the key implications of the license in simple terms like you did to me on the user list. That makes it much more likely that people will read through that section and grasp the key ideas you want to convey (of course, with a disclaimer that they don't replace the actual text of the license).

> Based on this, most new features will probably be added to "develop" first, but some bug fixes or simple improvements may get dropped directly into "master" first and then be merged forward into "develop".

Sounds good to me. I would add some summary of this to the contributing guide.

ronf:
Thanks again Nicholas for your work here!

I've incorporated the changes you suggested (with some edits mentioned above) into my develop branch. I have also added a new section to the contributing guide with information on the intended use of the master & develop branches.

nchammas:
Sounds good.

FYI: I noticed that some of the formatting seems off, though. Some code blocks show up as quote blocks instead, and the opening example is lacking [syntax highlighting](https://github.com/ronf/asyncssh/blob/34d8d39c9f6d1294075dd3acda51ef0c9637544b/README.rst#asyncssh-asynchoronous-ssh-for-python).

ronf:
Yeah, I've had some trouble with the way GitHub interprets restructured text. It doesn't seem entirely consistent with Sphinx, which is what I use the documentation and the only thing I can easily test the formatting with before I check in. I'll play around with it a bit and see if I can find something that works better for both use cases.

ronf:
I was able to fix the syntax highlighting for the code block, but the GitHub RST parser just doesn't seem to have any way to support indented blocks, at least not in a way that's also compatible with Sphinx. It wants to turn everything into a block quote. So, I think this is the best I can do for now.

PR #25, opened 2015-08-03
Turn all SFTPServer methods into coroutines. This allows asynchronous  implementation of the server.

jonathanslenders:
Hi @ronf,

This pull request makes some changes to the SFTP server. It allows all handlers to be coroutines.

The code as it was expected the server to host files from the local filesystem. In that case it makes sense to use blocking calls (for stat, fopen, read, write, fclose, etc...).

However, we needed an SFTP server that was asynchronous (that was why I used asyncio). The files that are served through the SFTP server did not come from a physical filesystem, but from an HTTP API. So, blocking is not an option.

Are you open for this kind of change? I think the power of asyncio is actually that we can do everything asynchronously. I'm not sure my pull request is 100% correct, you can have a look at it.

Cheers,
Jonathan

### DIFF
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -23,6 +23,7 @@ import time
 from collections import OrderedDict
 from fnmatch import fnmatch
 from os import SEEK_SET, SEEK_CUR, SEEK_END
+from asyncio.queues import Queue
 
 from .constants import DEFAULT_LANG
 
@@ -2777,6 +2778,16 @@ class SFTPServerSession(SFTPSession, SSHServerSession):
         self._file_handles = {}
         self._dir_handles = {}
 
+        self._packet_queue = Queue()
+
+        # Start process coroutine.
+        @asyncio.coroutine
+        def start_processing():
+            while True:
+                data = yield from self._packet_queue.get()
+                yield from self._process_packet_async(*data)
+        asyncio.async(start_processing())
+
     def _get_next_handle(self):
         """Get the next available unique file handle number"""
 
@@ -2833,7 +2844,10 @@ class SFTPServerSession(SFTPSession, SSHServerSession):
 
     def _process_packet(self, pkttype, pktid, packet):
         """Process other incoming SFTP packets"""
+        self._packet_queue.put_nowait((pkttype, pktid, packet))
 
+    @asyncio.coroutine
+    def _process_packet_async(self, pkttype, pktid, packet):
         try:
             if pkttype == FXP_EXTENDED:
                 pkttype = packet.get_string()
@@ -2844,7 +2858,7 @@ class SFTPServerSession(SFTPSession, SSHServerSession):
                                 'Unsupported request type: %s' % pkttype)
 
             return_type = self._return_types.get(pkttype, FXP_STATUS)
-            result = handler(self, packet)
+            result = yield from handler(self, packet)
 
             if return_type == FXP_STATUS:
                 result = UInt32(FX_OK) + String('') + String('')
@@ -2876,6 +2890,7 @@ class SFTPServerSession(SFTPSession, SSHServerSession):
 
         self.send_packet(Byte(return_type), UInt32(pktid), result)
 
+    @asyncio.coroutine
     def _process_open(self, packet):
         """Process an incoming SFTP open request"""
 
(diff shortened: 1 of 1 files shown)


ronf:
Hi @jonathanslenders,

Your use case definitely makes sense here, and I'd be happy to a take a closer look. When I first read through the SFTP spec, it talked about out-of-order handling of operations and some of the considerations there, where you could allow things like reads to happen in parallel but you needed to make sure that something like a write followed by a read would always happen in a fixed order if the ranges overlapped. I've been thinking about whether I wanted to support that sort of thing if I allowed async completion of the packet handlers, but hadn't worked out all the details yet.

From a quick look at your request, I see you're still serializing the requests in this case -- you're just doing it in a coroutine to allow async operations in the calls to the SFTPServer object. That's certainly a reasonable first step, and it could potentially be extended later on to support "safe" parallelism.

I have some other changes I'm in the middle of right now in the develop branch (related to some of the unit test code I've been adding), and I'd like to get some of that checked in before I pick up this work, but hopefully I can get to a good place there and then come back to this in the next week or two.

jonathanslenders:
Hi @ronf,

That's interesting, I didn't even know that out-of-order handling was possible. I kept the ordering indeed, that was something I learned from implementing the Redis protocol, where ordering is required.

Take the time that you need and keep up the good work!
Jonathan

jonathanslenders:
There was actually one more place where I'd have preferred a coroutine instead of a blocking method.
That's ``validate_password`` in ``asyncssh.SSHServer``. It's very hard right now to use a third-party authentication server. (E.g. ldap, oauth, etc...)

ronf:
Yeah, the question about supporting async operations in the auth callbacks came up once before on the asyncssh-users mailing list. Thanks for reminding me about it -- it was something I wanted to look into to see if I could add that support in a backward-compatible way.

I just checked in a change to the develop branch which adds support for the following SSHServer callbacks to optionally be defined as coroutines if they need to perform "blocking" operations:

        validate_public_key
        validate_ca_key
        validate_password
        get_kbdint_challenge
        validate_kbdint_response

I'm also going to look to see whether something like that can be done on the client side as well, and I haven't forgotten about your request about SFTP. I'll be tackling that soon as well.

If you get a chance to check out the change above, I'd love to hear if it works for you!

ronf:
Asynchronous client auth support is now available in the develop branch for the following SSHClient callbacks:

        public_key_auth_requested
        password_auth_requested
        password_change_requested
        kbdint_auth_requested
        kbdint_challenge_received

jonathanslenders:
Thanks! That's very cool.

jonathanslenders:
HI @ronf, I just tried these changes. In my case, ``validate_password`` is a coroutine now, and that works perfect! It makes the code a lot cleaner. Thanks.

ronf:
That's great to hear, Jonathan!

I just checked in the SFTP support for this in the develop branch, so all of the methods in SFTPServer can now optionally be coroutines. It's similar to your change, but to keep backward compatibility I haven't actually changed the parent class to automatically declare all of the methods to be coroutines. Instead, the packet processing code checks after calling each method whether it needs to yield from it or not.

I also added code to clean up the packet queue processing task when the SFTP server exits.

For now, I didn't add support for map_path() and reverse_map_path() to be coroutines, as I'm not sure how useful that is and it makes the code somewhat messy. I'm really not sure how useful that feature is in general, and may eventually remove or rewrite it. What do you think -- can you see a reason why those might need to be coroutines? Do you think name mapping like that is useful, or would it be better to just let subclasses do that themselves if they need it?

jonathanslenders:
Hi @ronf, thank you, that is nice!

About ``(reverse_)map_path``. I didn't use that at all. Actually they are not even present in my own ``SFTPServer`` class (which is inheriting from object). So,  a bit confusing indeed.

ronf:
Yeah - that's one of the things I don't really like about the current implementation. The subclass of SFTPServer you create has to explicitly know when to call map_path() and reverse_map_path() if you override any of the other methods and you still want them to work. I could have put those calls in SFTPServerSession, but I really didn't want to dictate that mapping like this always had to be done. So, maybe having this support in the parent class isn't the right approach.

Back to the topic of this pull request, let me know if the version in develop meets your needs. If it does, I'll make sure to include it in the next merge I do to master.

PR #26, opened 2015-08-10
Accept WRITE flag in 'open' sftp call. 

jonathanslenders:
The sshfs tools (using Ubuntu) sends the WRITE flag only when creating/copying files.
I literally receive the value ``2``.

First, ``SFTPError(FX_FAILURE, 'Unsupported open flags')`` was raised.

### DIFF
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -2757,6 +2757,9 @@ class SFTPServerSession(SFTPSession, SSHServerSession):
 
     _open_modes = {
         FXF_READ:                                      'rb',
+        FXF_WRITE:                                     'wb',
+        FXF_WRITE | FXF_CREAT:                         'wb',
+        FXF_WRITE | FXF_EXCL:                          'xb',
         FXF_WRITE | FXF_CREAT | FXF_TRUNC:             'wb',
         FXF_WRITE | FXF_CREAT | FXF_APPEND:            'ab',
         FXF_WRITE | FXF_CREAT | FXF_EXCL:              'xb',


ronf:
When sshfs passes just FXF_WRITE, what is the expected behavior in the case where the file doesn't previous exist? Will it create it, or does it expect to get an error back? Similarly, when a file already exists, does it truncate the file on open, or does it keep the existing contents intact? If you write less than the previous length of the file, do you end up with a mix of old and new data, or would only the new data be present in the file after that write?

As you've defined it here, passing FXF_WRITE by itself or FXF_WRITE | FXF_CREAT is treated the same as passing in all three flags FXF_WRITE | FXF_CREAT | FXF_TRUNC, meaning it would create the file if necessary, or truncate it if it already existed. However, from what I can see from the OpenSSH SFTP server implementation, that's not what would happen if you passed just FXF_WRITE to it. It looks to me like passing FXF_WRITE by itself (without FXF_CREAT) would not create a file if it didn't already exist, and would not truncate it if it did exist, though it would only be open for write and not read/write. 

Also, you added a case of FXF_WRITE | FXF_EXCL without FXF_CREAT, but I don't think that's legal. according to the SFTP version 3 spec:

    SSH_FXF_EXCL
        Causes the request to fail if the named file already exists.
        SSH_FXF_CREAT MUST also be specified if this flag is used.

So, passing in FXF_EXCL without FXF_CREAT probably should be an error. The same applies for passing in FXF_TRUNC without FXF_CREAT.

Unfortunately, supporting opening for write without create or truncate is not supported in the native Python open() modes, but I believe I could get the necessary behavior here using the more primitive os.open() flags via the "opener" argument to open(). If you can confirm the expected behavior for me using something like sshfs talking to OpenSSH, I'd be happy to attempt this change.

jonathanslenders:
Some obvervations when mounting through sshfs:

- running "echo hello > world" opens in mode ``26 (=hex 1a) = FXF_TRUNC|FXF_CREAT|FXF_WRITE``
- running "cp oldfile newfile_in_mount" opens in mode ``42  (=hex 2a)  = FXF_EXCL|FXF_CREAT|FXF_WRITE``
- running "touch newfile" opens in mode ``10 (=hex a) = FXF_TRUNC``
- running "touch existing_file" opens in mode`` 2  =FXF_WRITE``

- running "vim a_new_file" opens a swapfile in mode ``43 (=hex 2b) = FXF_EXCL|FXF_CREAT|FXF_WRITE|FXF_READ``
-running "vim existing_file" opens the existing file in mode ``1 = FXF_READ`` and a swapfile in mode ``43``.

ronf:
Mode 26 (0x1a) looks good and is supported now. The same goes for modes 1, 42 (0x2a) and 43 (0x2b).

Mode 10 (0x0a) is actually FXF_WRITE | FXF_CREAT, not FXF_TRUNC (which wouldn't be legal by itself). That's not supported currently, but I can try to support it once I confirm the behavior on an existing file. Since it is used by "touch", I'm guessing it doesn't truncate the file if it already exists, but it also doesn't return an error in that case. That means neither the 'wb' or the 'xb' open modes would be right for that case. I'll have to override the normal Python open() flags to handle this case.

Regarding mode 2 for the 'echo >> existing_file' on an existing file, can you double-check that? It looks like that should be an append, but perhaps sshfs is handling that itself after the open. Are you seeing a separate request to seek to the end of the file after it is opened? I'm guessing that will be the case. Once again, none of the existing Python open modes work for this case where you want to open write-only, want an error if the file doesn't exist, and don't want to truncate the file if it does exist. However, I may be able to do it with an override.

One challenge here is going to be around backward compatibility. My current API only passes the string version of the open mode to the SFTPServer open() method. So, I'm either going to need to make up new strings for these two new cases or I'm going to move the mapping into SFTPServer and pass the raw flags through. Either way someone who overrode open() would probably need to change their implementation to properly handle the new modes, so I think I'm better off doing the latter to allow arbitrary combinations.

For consistency, I'll probably allow the raw flags on the SFTPClient as well, but there I can support both the string version and the numeric version easily so there's no backward compatibility problem there.

ronf:
Ok - I just checked in a change to develop that should resolve this issue. Both SFTPClient.open() and SFTPServer.open() have been changed to support passing in a bitmask of AsyncSSH pflags values so that all open modes can be supported. Not all combinations make sense, so some may lead to errors later if you try to do something like a write when you've opened for read+append, but other combinations that were previously unsupported now work, including all of the ones listed above that SSHFS was using.

Other than an argument name change from "mode" to "pflags_or_mode", the SFTPClient.open() call is backward compatible, as you can still pass in a Python-style mode string rather than the raw flag values if you want to. However, the SFTPServer.open() callback will now always be given the raw pflags, so any application code which implemented that will need to be updated.

The default implementation of SFTPServer.open() automatically translates the SSH pflags to local OS open flags and passes them into os.open(). So, all reasonable combinations of these flags should now be supported even if the higher level Python open() call had no way to directly request that combination of flags.

PR #33, opened 2015-10-29
Fix a bug in keyboard-interactive auth

zenbot:
When fetching the password during keyboard interactive auth we did not handle
the case of SSHClientConnection.password_auth_requested being a coroutine,
meaning that where we were expecting a password string we would under some/all
circumstances have a generator instead.


ronf:
Thanks very much for the report! This fix is now checked into both develop (86102f1) and master (7131e90).

Issue #47, opened 2016-02-05
Reading `~/.ssh/known_hosts` runs on every connection


qix:
The library I've been working on connects to a large number of hosts simultaneously which seems to break the asyncssh `known_hosts` reading implementation. As far as I can tell it's designed to scan through and decode the entire file for every connection -- which causes obvious problems.

I don't know if you want to try cache the scan internally, or if you'd rather I passed in the `known_hosts` variable into `create_connection`... at which point the `known_hosts._parse_entries` function should be properly exposed.

ronf:
If you are connecting to the same host repeatedly and you know the known hosts list isn't changing, my recommendation would to be make your own call to match_known_hosts() to get back the keys and then store those on a per-host basis. On each new connection, you can pass in that tuple of (host_keys, ca_keys, revoked_keys) in place of a file name or known hosts byte string and AsyncSSH will use them without doing the matching again.

It would be difficult to automatically cache this information. A separate result would have to be cached for each host/addr/port combination that is passed in. Also, since the file can change at any time, it would be necessary to read the file (or at least do something like a stat to check the timestamp) to know when the cache needed to be invalidated. In the end, I'm not sure there would be enough value out of doing the caching to justify it in the typical use case.

qix:
I don't think using `match_known_hosts` will help. The problem I'm experiencing is due to `match_known_hosts` being far too slow to run on every `host/port` pair.

Just submitted pull request #48 which seperates the parsing so that it only needs to be done once.

ronf:
Ok - I see what you're looking for now, but I must admit I'm a bit surprised that the parsing operation was take a noticeable amount of time compared to the matching. Have you done any timing tests on this, or could you provide a sample known_hosts file I could use to explore this a bit more? Since the match function is where it calls things like fnmatch() multiple times on each entry and it scans all of the entries linearly every time, I would have expected that to be where most of the time went.

That said, I'm happy to consider a split like this, but I'd probably name things a bit differently to be more consistent with the rest of the API. I'm thinking of having import_known_hosts() and read_known_hosts() functions to parse known hosts data from a string and a file, respectively. Both of these would return an SSHKnownHosts object with a match() method that would do the matching against the parsed data. I'd then have match_known_hosts() allow create_connection()'s known_hosts argument to take an SSHKnownHosts as another option, in which case it would use the parsed data and only do the match call.

[...]

PR #48, opened 2016-02-07
Seperate known_hosts parsing/matching

qix:
See issue #47 

I'm good with the Eclipse Public License v1.0 and however you wish to manage a contributors list.


ronf:
Take a look at commit 133ce51 and let me know if this provides what you're looking for.

With this, you can call import_known_hosts() or read_known_hosts() to load the pattern data, and then pass the value returned (an SSHKnownHosts object) into create_connection() via the known_hosts argument instead of a filename.

PR #56, opened 2016-05-30
important issue: to append a general object to factory instance once it is created

ghostplant:
important issue: to append a general object to session factory instance once it is created synchronizedly,

otherwise, "asyncssh.SSHClientSession.data_received" might get access to a object which haven't been setting the expected value:

...
...
  def data_received(self, data, datatype):
    self._websocket.send(data)                    ## Line A
...
...

...
...
  chan, session = yield from conn.create_session(MySSHClientSession, 'bash', term_type='xterm', term_size=(80, 24))
  session._websocket = websocket_map[userid]      ## Line B
...
...

Line A might run before Line B, making error exception for a certain possiblity

====================================

After using this patch, it will work well and no more exception

...
...
  def session_started(self):
    extract_data = self._factory_obj_append
    self._websocket = extract_data[0]

  def data_received(self, data, datatype):
    self._websocket.send(data)
...
...

...
...
  conn, client = yield from asyncssh.create_connection(MySSHClient, hostname, username=username, port=int(s_port), password=password, known_hosts=None, factory_obj_append=(_websocket, ..))
..
  chan, session = yield from conn.create_session(MySSHClientSession, 'bash', term_type='xterm', term_size=(80, 24))
...
...

====================================

This feature is useful to implement websocket + asyncssh = webssh

Signed-off-by: cuiwei13 <cuiwei13@pku.edu.cn>

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -790,7 +790,7 @@ class SSHClientChannel(SSHChannel):
 
     @asyncio.coroutine
     def create(self, session_factory, command, subsystem, env,
-               term_type, term_size, term_modes, agent_forwarding):
+               term_type, term_size, term_modes, agent_forwarding, factory_obj_append):
         """Create an SSH client session"""
 
         packet = yield from self._open(b'session')
@@ -799,6 +799,7 @@ class SSHClientChannel(SSHChannel):
         packet.check_end()
 
         self._session = session_factory()
+        self._session.factory_obj_append = factory_obj_append
         self._session.connection_made(self)
 
         for name, value in env.items():
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -1941,7 +1941,7 @@ class SSHClientConnection(SSHConnection):
     def __init__(self, client_factory, loop, kex_algs, encryption_algs,
                  mac_algs, compression_algs, rekey_bytes, rekey_seconds,
                  host, port, known_hosts, username, password, client_keys,
-                 agent, agent_path, auth_waiter):
+                 agent, agent_path, auth_waiter, factory_obj_append=None):
         super().__init__(client_factory, loop, kex_algs, encryption_algs,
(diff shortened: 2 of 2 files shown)


ronf:
There are other ways to do this that don't require a patch to AsyncSSH. The easiest is probably to define a session factory method that passes in the extra argument to your session subclass constructor. For instance, here's a slightly modified version of the "simple client" example:

```Python
class MySSHClientSession(asyncssh.SSHClientSession):
    def __init__(self, websocket):
        self._websocket = websocket

    '''Methods to process data on the session go here'''

@asyncio.coroutine
def run_client():
    '''Add code here to create a web socket and assign it to local variable websocket'''

    def session_factory():
        return MySSHClientSession(websocket)

    with (yield from asyncssh.connect('localhost')) as conn:
        chan, session = yield from conn.create_session(session_factory, 'ls abc')
        yield from chan.wait_closed()

    yield from conn.wait_closed()

try:
```

You can add as many arguments as you like to the MySSHClientSession constructor and the session_factory() function you define can access whatever local variables it likes when passing values to your session object.

ghostplant:
Hi, it is a good idea but doesn't work.

if session_factory method is defined as:

    websocket = ...
    def session_factory():
        return MySSHClientSession(websocket)

then program throws an exception:

  File "/usr/lib/python3/dist-packages/asyncssh/connection.py", line 2105, in create_session
    env, term_type, term_size, term_modes))
  File "/usr/lib/python3/dist-packages/asyncssh/channel.py", line 760, in create
    self._session = session_factory()
  File "./webssh-tunnel.py", line 42, in ssh_session_factory
    return MySSHClientSession(_websocket)
TypeError: object() takes no parameters


However, if outer parameter is dropped, making session_factory method defined as:

    def session_factory():
        return MySSHClientSession()

Then there would be no exception. But websocket parameter still cannot append to the factory class instance.

PR #63, opened 2016-09-30
tests: discard ssh-agent error output

vincentbernat:
ssh-agent being a setgid program, it is possible the linker complain
about not being able to preload a library specified in `LD_PRELOAD`
environment variable. This can happen when people are using `eatmydata`,
a program disabling `sync()` and friends to speed up compilation.

### DIFF
--- a/tests/server.py
+++ b/tests/server.py
@@ -110,7 +110,7 @@ class ServerTestCase(AsyncTestCase):
                                                       cls._server_port))
         run('cat skey.pub >> .ssh/known_hosts')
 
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for this, Vincent!

I'm in the middle of making some other changes to the agent code right now in the develop branch to add the remaining operations such as adding and removing keys. I'll include the changes you have proposed here as part of that commit.

ronf:
Vincent, the changes you have proposed here are now merged into both the master and develop branches, and part of the new 1.7.1 release. Thanks again!

PR #66, opened 2016-11-18
Add another SkipTest for ssh-agent in tests/server.py

cooperlees:
Got the tests to pass now.

- import only the parts of the modules we need
- Add a SkipTest in server.py where we try to execute ssh-agent
- Be consistent and do partial imports in test_agent.py too

### DIFF
--- a/tests/server.py
+++ b/tests/server.py
@@ -18,6 +18,8 @@ import signal
 
 import asyncssh
 from asyncssh.misc import async_context_manager
+from subprocess import CalledProcessError
+from unittest import SkipTest
 
 from .util import run, AsyncTestCase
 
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for catching this additional case of running ssh-agent, Cooper. I'd prefer not to skip the test completely in this case, though, as a large number of test cases rely on this ServerTestCase class and many of those tests don't rely on ssh-agent. I've made an alternate fix that allows ServerTestCase initialization to complete successfully even when ssh-agent isn't available, with a function available to check on this so that only tests which need ssh-agent are skipped. I've identified these tests (about 11 of them) and added the necessary code to make them conditional.

Regarding the imports, my rule of thumb is to generally import the top-level package rather than specific items from the package to avoid potential name conflicts. I make an exception for certain things like datetime.datetime or fnmatch.fnmatch, and also for some classes in collections and some constants, but otherwise that's the rule I try to follow everywhere, using "from" imports only for local stuff needed in the tests that's not part of AsyncSSH's public API (and thus not available from a top-level import).

The updated fix is available in commit 05df291. Hopefully this one will work for you! I really appreciate your help in getting this right.

PR #67, opened 2016-12-05
add read_until method for SSHReader

wwjiang:


ronf:
Thank you for your contribution. There are a few minor issues here that I'll need to correct, but I'd be happy to add this to AsyncSSH. Specifically:

- The function like this added in Python  3.5.2's StreamReader class is named "readuntil", not "read_until", and its argument is called "separator". I'm going to mimic that API when adding it to my SSHReader class.

- There are some small differences between readline() and readuntil() in terms of EOF handling. Specifically, readuntil() can raise IncompleteReadError if EOF is received before the separator appears, whereas readline() will return the partial line or an empty string in this case. While it makes sense to implement readline() using readuntil(), that difference will need to be preserved.

- With AsyncSSH, the separator should be specified as a byte string when encoding on the channel is set to None, rather than always taking a string and converting it using the default system encoding in that case.

I'm in the middle of some other code changes right now, but I'll try and get this into my "develop" branch  as soon as I've finished with those.

ronf:
I've just checked an implementation of readuntil() for SSHReader into the "develop" branch. I'd appreciate any feedback you have on it if you are able to test it there. As noted in the README, you can do an install of the develop branch using the following 'pip' command:

    pip install git+https://github.com/ronf/asyncssh@develop

I still have a couple of additional work items I'd like to finish before I roll out a new release, but I expect this to be available in an official release in the next week or two.

Thanks again for the suggestion!

ronf:
This change is now available in AsyncSSH release 1.8.0. Thanks again for your suggestion!

PR #78, opened 2017-02-12
Don't use "async" as argument name

jwilk:
The use of `async` and `await` as names was deprecated in Python 3.6.
They will become become reserved keywords in Python 3.7.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


ronf:
Thanks for reporting this, Jakub. Commit 64efc83 in the "develop" branch fixes this, and will be rolled into "master" in the next release.

PR #88, opened 2017-04-30
Fix SSH connection with known hosts object.

viotti:
The connection was failing when the `known_hosts` parameter was a `SSHKnownHosts` object previously imported from a string by calling `import_known_hosts()`, as documented [here](http://asyncssh.readthedocs.io/en/latest/api.html#specifying-known-hosts).

Python was throwing an exception since the `SSHKnownHosts` object was not "unpackable". Using the `self._known_hosts.match` method, which returns the proper tuple, appears to fix the error.

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -1966,10 +1966,20 @@ class SSHClientConnection(SSHConnection):
                 server_host_keys, server_ca_keys, revoked_server_keys = \
                     match_known_hosts(self._known_hosts, self._host,
                                       self._peer_addr, self._port)
-            else:
+
+            # Duck typing SSHKnownHosts.
+            elif hasattr(self._known_hosts, 'match'):
                 server_host_keys, server_ca_keys, revoked_server_keys = \
                     self._known_hosts.match(self._host, self._peer_addr,
                                             self._port)
+
+                server_host_keys = load_public_keys(server_host_keys)
+                server_ca_keys = load_public_keys(server_ca_keys)
+                revoked_server_keys = load_public_keys(revoked_server_keys)
+
+            else:
+                server_host_keys, server_ca_keys, revoked_server_keys = \
+                    self._known_hosts
 
                 server_host_keys = load_public_keys(server_host_keys)
                 server_ca_keys = load_public_keys(server_ca_keys)
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -1968,7 +1968,8 @@ class SSHClientConnection(SSHConnection):
                                       self._peer_addr, self._port)
             else:
                 server_host_keys, server_ca_keys, revoked_server_keys = \
-                    self._known_hosts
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for contributing this. You're right that passing in a previously loaded known_hosts object isn't handled correctly right now, as the explicit test for str & bytes prevents match_known_hosts from being called. The intention was that match_known_hosts would handle all of these cases except for where trusted host keys, trusted CA keys, and revoked keys is directly provided and no match call is needed.

I'll check in a variation of this fix in the develop branch soon, once I get some other outstanding changes committed.

viotti:
@ronf thank you for this library. I've been using Spur/Paramiko for quite some time now, which are great projects BTW, but I wish I had found this sooner. I have a web project which must leverage a GNU Make toolchain to build ARM firmware, and the combination of aiohttp and asyncssh fits like a glove. No contention on a web request while the compiler is running on another via a sub-process.

Im curious about that matching logic you mentioned. In the case the keys are explicitly provided as a tuple of lists, why no match call is needed? What if the wrong set of host keys are provided? The connection will be dropped anyway at some point later?

ronf:
I'm very happy to hear that AsyncSSH is working well for you!

You can think of the case where keys are provided explicitly as being the equivalent of the caller doing whatever was needed to parse the known_hosts file and pull out the right keys in advance of requesting the connection. In other words, they are providing the host and port to connect to, and they might know exactly what keys to trust to connect to that host. When this is the case, it's a lot faster to directly provide the keys than it would be to make AsyncSSH walk through the entire known_hosts file looking for that host/address/port.

Regarding the fix for this bug, I'm thinking I might expand it to also accept a callable which accepts a host, addr, and port and returns the tuple of lists of trusted keys. That way, if a caller wanted to provide their own match function which was completely independent of SSHKnownHosts, they'd be able to do so by simply passing that function in as a "known_hosts" argument, rather than having to know to create their own class with a "match" method defined on it. The other forms (str, bytes, SSHKnownHosts, or tuple of key lists) would also continue to be accepted.

I'm debating about whether to also support the match function provided this way being a coroutine, so the caller could actually block in the match code if they needed to. This would allow them to do something like do a network lookup to get what keys to trust. It's a bit ugly as this is called from an asyncio callback which isn't a coroutine right now, but I ran into similar issues when I added support the auth methods to be coroutines and there are ways to make it work.

ronf:
I just checked in a fix for this in the "develop" branch, and also added the ability to provide your own matching function as an argument to known_hosts, in addition to the previous options to provide a known_hosts file, known_hosts data as a byte string, a previously loaded SSHKnownHosts object, or a tuple of lists of keys.

The new matching function can return the same sort of values supported when specifying the tuple of key lists, meaning the lists can refer to keys by file name or provide them as byte strings and AsyncSSH will take care of reading/importing them as needed.

I should be able to put out a release with this fix over the weekend.

ronf:
This fix and enhancement is now available in AsyncSSH version 1.10.0.

PR #89, opened 2017-05-01
Fix python3 highlighting in docs

ZhukovAlexander:


ronf:
The Sphinx documentation is somewhat inconsistent about this. According to the highlight_language docs:

> The default language to highlight source code in. The default is 'python3'. The value should be a valid Pygments lexer name, see Showing code examples for more details.
>
> New in version 0.5.

However, it goes on to say:

> Changed in version 1.4: The default is now 'default'. It is similar to 'python3'; it is mostly a superset of 'python'. but it fallbacks to 'none' without warning if failed. 'python3' and other languages will emit warning if failed. If you prefer Python 2 only highlighting, you can set it back to 'python'.

This isn't entirely consistent with the change log itself, though, which for version 1.4 says:

> The default highlight language is now Python 3. This means that source code is highlighted as Python 3 (which is mostly a superset of Python 2), and no parsing is attempted to distinguish valid code. To get the old behavior back, add highlight_language = "python" to conf.py.

Is there something specific which wasn't highlighting correctly that caused you to propose this?

ZhukovAlexander:
> Is there something specific which wasn't highlighting correctly that caused you to propose this?

Yes, without this change, the `async-await` syntax is not highlighted. For comparison, here is the version of docs built with this change: http://asyncssh-azhukov-fork.readthedocs.io/en/latest/#simple-client, as opposed to the current version: http://asyncssh.readthedocs.io/en/stable/#simple-client

ronf:
Ah, thanks! I knew this was a problem, but the last time I looked over at ReadTheDocs, they didn't have support for this yet.

PR #92, opened 2017-05-12
Make sure the `exit` method of the SFTPServer is called at the end of an SCP session

arthurdarcet:
Everything is in the title!

### DIFF
--- a/asyncssh/scp.py
+++ b/asyncssh/scp.py
@@ -379,6 +379,7 @@ class _SCPSource(_SCPHandler):
             self.handle_error(exc)
         finally:
             self.close()
+            yield from self._fs.exit()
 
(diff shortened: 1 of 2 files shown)


ronf:
Thank you for submitting this. I can confirm the problem. There are some minor issues with the proposed fix, though, such as an argument being passed to the SFTPServer's exit() function that shouldn't be there. Also, I'd like to avoid having to touch the LocalFile and SFTPServerFile classes to fix this.

See my commit 57df31a in the "develop" branch for an alternate fix. This change will be rolled into the next release.

ronf:
This is now released in version 1.10.1. Thanks again for finding this issue and suggesting a fix!

PR #95, opened 2017-06-07
forward passphrase to agent.get_keys()

obeleh:
This fixes an issue for me where I tried to supply the passphrase but it wasn't used

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -4358,7 +4358,7 @@ def create_connection(client_factory, host, port=_DEFAULT_PORT, *,
             agent = yield from connect_agent(agent_path, loop=loop)
 
             if agent:
-                client_keys = yield from agent.get_keys()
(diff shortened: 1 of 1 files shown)


ronf:
I don't understand - the get_keys() method does not accept a passphrase argument, and doesn't need one. The private keys are kept in the agent and the passphrase (if needed) would have been provided when the keys were loaded by the agent. It is not needed when the SSHClientConnection queries the agent for the list of available keys.

obeleh:
You're right. I copied the change wrong. Let me fix it

obeleh:
Just ignore the PR. I see it was fixed already. Sorry for bothering you

ronf:
No worries. Was this the issue in the non-agent case where the passphrase wasn't being used when reading the default client keys (69f1524)?

obeleh:
yes

PR #123, opened 2018-02-02
Do not throw exception when user doesn't specify x509_trusted_certs

termim:
Check if default ~/.ssh/ca-bandle.crt file exists before trying to load it.
Otherwise load_certificates() throws FileNotFoundError which clogs the logs.

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -4482,11 +4482,12 @@ def create_connection(client_factory, host, port=_DEFAULT_PORT, *,
                        signature_algs, x509_trusted_certs is not None)
 
     if x509_trusted_certs is ():
-        try:
-            x509_trusted_certs = load_certificates(
-                os.path.join(os.path.expanduser('~'), '.ssh', 'ca-bundle.crt'))
-        except OSError:
-            pass
+        ca_bandle = os.path.join(os.path.expanduser('~'), '.ssh', 'ca-bundle.crt')
+        if os.path.exists(ca_bandle):
+            try:
+                x509_trusted_certs = load_certificates(ca_bandle)
+            except OSError:
+                pass
     elif x509_trusted_certs is not None:
         x509_trusted_certs = load_certificates(x509_trusted_certs)
 


ronf:
I don't understand this change. The whole point of the try/except looking for OSError is to make sure we can silently move forward when the file doesn't exist or it can't be read due to something like permissions. Which "logs" are being written to here when this error occurs? There's no logging of these errors in AsyncSSH itself, and the exception should not be visible outside of AsyncSSH.

Also, the proposed change would introduce a race condition. Just because the file exist call returns "True" doesn't mean the file will still exist when load_certificates() is called. With exception handling code like this, you generally want to just perform the operation and deal with whatever errors occur. That's why it is written the way it is now.

termim:
You are right, that happens only in a debug session in my IDE. When I debug some script which has a long list of hosts to connect to it quickly becomes very annoying. But I found a workaround - I pass x509_trusted_certs = None to connect and it is silent even in a debug mode.

As for the race condition I guess the file could be deleted right after it was read and before parsed and used, certificate could became untrusted in the middle etc., so it is not very important IMHO.

Anyway, closing it as there is a workaround for me. Sorry for the noise and thanks for the great project!

ronf:
Glad you were able to find a workaround -- it makes sense that passing in that arg would prevent it from trying to load anything from the default location.

If you don't mind my asking, I'm still curious about where the logging is coming from, though, since I don't see how AsyncSSH itself would be generating these messages. Is it something your IDE is doing whenever file accesses fail, even if the code is trying to catch the exception? What do the messages look like? Perhaps that'll give a clue as to where they're coming from.

termim:
That is actually in documentation, my bad I missed it - so many options. As I see now it is probably hard to choose defaults that would satisfy every use case. The good thing it is all adjustable through parameters.

When script is running under debugger my IDE (I use eric6) reports exceptions even if they are handled in the code. There is a way to mask them though.

ronf:
Ah, yeah - the IDE reporting exceptions even when they're handled in the code would explain what you saw. There are some other places you might see this as well, such as when AsyncSSH tries to load the default key files (.ssh/id_dsa, .ssh/id_rsa, etc.). You could manually specify the client keys to load to address that if you needed to, though.

Thanks for taking the time to explain, and I'm glad to hear you were able to work around this. If you run into other issues, feel free to post here any time!

Issue #112, opened 2017-11-09
created process can't be terminated


gsauthof:
Consider this small example:

    #!/usr/bin/env python3
    
    import asyncio
    import asyncssh

    hostname = 'example.org'
    user = 'juser'
    
    async def f():
      async with asyncssh.connect(hostname, username=user) as c:
          async with c.create_process('sleep 300') as p:
            print('sleep')
            await asyncio.sleep(3)

            print('terminate')
            p.terminate()

            print('wait')
            await p.wait()
    
            print('done')

    if __name__ == '__main__':
      loop = asyncio.get_event_loop()
      loop.run_until_complete(f())
      loop.close()

Expected output:

    sleep
    terminate
    wait
    done
    (program terminates - total runtime: 3 seconds or so)

Actual output:

    sleep
    terminate
    wait
    (program is still running)

That means it seems that the TERM signal isn't delivered.

(When I log into the remote system I see the process still running in `pstree` output - also after I kill the test program I see the orphaned sleep process via `pgrep -u juser -P 1 -l -f`)

I tested this against an sshd on a remote CentOS 7, an sshd on a remote Solaris 10 and a local sshd on Fedora 25 (localhost). Same results each time.

I also switched the `'sleep'` command with `'/usr/bin/sleep'`, and the `p.terminate()` call with `p.kill()` and `p.send_break(1000)`. Doesn't make a difference.

Please advise if I'm using the asyncssh in a wrong way or if I'm missing something obvious.

ronf:
This came up recently on the asyncssh-users mailing list. Are you using OpenSSH as your SSH server? If so, the problem is that OpenSSH actually doesn't implement the "signal" channel request. As a result, even though AsyncSSH is sending the proper message listed in the SSH RFC, nothing happens when the send_signal(), terminate(), and kill() methods are called. You can find more info at https://bugzilla.mindrot.org/show_bug.cgi?id=1424. Unfortunately, this is a very old bug and as far as I know the patch listed there has never made it into a standard release.

It’s a bit of a hack, but depending on the signal you want to send it might be possible to do it by writing to stdin instead. If you set up the command to run in a PTY (by setting term_type), you can then write something like ‘\x03’ (Ctrl-C) to stdin to get the equivalent of a SIGINT, or ‘\x1c’ to get the equivalent of a SIGQUIT. I don’t know of anything you can write to get other signals like SIGTERM, SIGHUP, or SIGKILL, though.

[...]

PR #136, opened 2018-04-11
make note on openssh more visible

parmentelat:
related to #112

make note, about send_signal not being honored by openssh, more visible from kill() and terminate() for stupid readers like me


ronf:
Thanks for your contribution!

PR #146, opened 2018-05-22
Bug remote listener doesn't close

### LINKED ISSUE #145: BUG: Closing the remote forwarding server doesn't close the server
The SSHListener doesn't close a remote server.
The problem is in `SSHClientConnection.close_client_tcp_listener`.

Currently: 
```
yield from self._make_global_request(
b'cancel-tcpip-forward', String(listen_host), UInt32(listen_port))
```
Should be:
```
yield from self._make_global_request(
b'cancel-tcpip-forward', listen_host.encode(), UInt32(listen_port))
```
I will fix the issue and create a new pull request.

Itayazolay:
Closes #145 
with a binary SSH server the fix works, but with the current asyncssh implementation it doesn't work.
I've looked at the  [SSH RFC](https://tools.ietf.org/rfc/rfc4254.txt) and according to page 16, the local host should be passed as a string(bytes, in python 3).
Currently the `String` method return something like this:
```
In [3]: String("127.0.0.1")
Out[3]: b'\x00\x00\x00\t127.0.0.1'
```
When I checked it against my ssh server it didn't work(both openssh and dropbear implementations), but with my fix it did work.
I see that the "tcpip-forward" is working, but I don't really understand how.

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -2445,7 +2445,7 @@ class SSHClientConnection(SSHConnection):
         """Close a remote TCP/IP listener"""
 
         yield from self._make_global_request(
-            b'cancel-tcpip-forward', String(listen_host), UInt32(listen_port))
+            b'cancel-tcpip-forward', listen_host.encode(), UInt32(listen_port))
 
         self.logger.info('Closed remote TCP listener on %s',
                          (listen_host, listen_port))
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -660,16 +660,17 @@ class _TestTCPForwarding(_CheckForwarding):
         with (yield from self.connect()) as conn:
             listener = yield from conn.forward_remote_port('', 0, '',
                                                            server_port)
-
+        
             yield from self._check_local_connection(listener.get_port())
-
+            used_port = listener.get_port()
             listener.close()
             yield from listener.wait_closed()
             # Verify the port is available again.            
-            listener = yield from conn.forward_remote_port('', 0, '',
+            listener = yield from conn.forward_remote_port('', used_port, '',
                                                            server_port)
-            
-            yield from self._check_local_connection(listener.get_port())
+            self.assertIsNotNone(listener)
+            yield from listener.close()
+            yield from listener.wait_closed()
 
         yield from conn.wait_closed()
 
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -665,7 +665,9 @@ class _TestTCPForwarding(_CheckForwarding):
             used_port = listener.get_port()
             listener.close()
             yield from listener.wait_closed()
+            
             # Verify the port is available again.            
+            yield from asyncio.sleep(0.5)
             listener = yield from conn.forward_remote_port('', used_port, '',
                                                            server_port)
             self.assertIsNotNone(listener)
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -649,7 +649,33 @@ class _TestTCPForwarding(_CheckForwarding):
             self.assertIsNone(listener)
 
         yield from conn.wait_closed()
+    
+    @asynctest
+    def test_forward_remote_port_closed(self):
+        """Test forward remote port closes currectly."""
+        server = yield from asyncio.start_server(echo, None, 0,
+                                                 family=socket.AF_INET)
+        server_port = server.sockets[0].getsockname()[1]
 
+        with (yield from self.connect()) as conn:
+            listener = yield from conn.forward_remote_port('', 0, '',
+                                                           server_port)
+
+            yield from self._check_local_connection(listener.get_port())
+
+            listener.close()
+            yield from listener.wait_closed()
+            # Verify the port is available again.            
+            listener = yield from conn.forward_remote_port('', 0, '',
+                                                           server_port)
+            
+            yield from self._check_local_connection(listener.get_port())
+
+        yield from conn.wait_closed()
+
+        server.close()
+        yield from server.wait_closed()
+        
     @asynctest
     def test_cancel_forward_remote_port_invalid_unicode(self):
         """Test canceling TCP/IP forwarding with invalid Unicode in host"""
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -668,10 +668,10 @@ class _TestTCPForwarding(_CheckForwarding):
             
             # Verify the port is available again.            
             yield from asyncio.sleep(0.5)
-            listener = yield from conn.forward_remote_port('', used_port, '',
-                                                           server_port)
+            new_server = yield from asyncio.start_server(echo, None, used_port)
+
             self.assertIsNotNone(listener)
-            yield from listener.close()
+            new_server.close()
             yield from listener.wait_closed()
 
         yield from conn.wait_closed()


ronf:
I appreciate your effort here, but I think there's something wrong in your test environment. There's no way that this fix could work. As you said, the SSH RFC defines this argument as a string:

      byte      SSH_MSG_GLOBAL_REQUEST
      string    "cancel-tcpip-forward"
      boolean   want reply
      string    address_to_bind (e.g., "127.0.0.1")
      uint32    port number to bind

If you look at RFC 4251, you'll see that strings are defined as being prefixed with a 32-bit length:

   string

      Arbitrary length binary string.  Strings are allowed to contain
      arbitrary binary data, including null characters and 8-bit
      characters.  They are stored as a uint32 containing its length
      (number of bytes that follow) and zero (= empty string) or more
      bytes that are the value of the string.  Terminating null
      characters are not used.

      Strings are also used to store text.  In that case, US-ASCII is
      used for internal names, and ISO-10646 UTF-8 for text that might
      be displayed to the user.  The terminating null character SHOULD
      NOT normally be stored in the string.  For example: the US-ASCII
      string "testing" is represented as 00 00 00 07 t e s t i n g.  The
      UTF-8 mapping does not alter the encoding of US-ASCII characters.

So, the encoding you listed above for String("127.0.0.1") is correct.

I did a test here with the currently released AsyncSSH, and the port close function works just fine without this change. I even expanded my unit test to cover this (similar to what you propose here), and confirmed that I got an error trying to listen twice on a port without closing first, but could close the listener and bind to the same port again after that without a problem. Here's my diff for the unit test change:

```patch
index ad95491..cfe2fde 100644
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -576,6 +576,21 @@ class _TestTCPForwarding(_CheckForwarding):
 
             yield from self._check_local_connection(listener.get_port())
 
+            listen_port = listener.get_port()
+            listener2 = yield from conn.forward_remote_port('', listen_port,
+                                                            '', server_port)
+
+            self.assertIsNone(listener2)
+
+            listener.close()
+            yield from listener.wait_closed()
+
+            yield from asyncio.sleep(0.2)
+            listener = yield from conn.forward_remote_port('', listen_port,
+                                                           '', server_port)
+
+            self.assertIsNotNone(listener)
+
             listener.close()
             yield from listener.wait_closed()
 
```

Here's a portion of the log output showing the result of this:

```
INFO:asyncssh:[conn=0] Creating remote TCP forwarder from dynamic port to port 56626
INFO:asyncssh:[conn=0] Creating remote TCP listener on dynamic port
INFO:asyncssh:[conn=1] Creating local TCP forwarder on dynamic port
DEBUG:asyncssh:[conn=1] Assigning dynamic port 56628
INFO:asyncssh:[conn=1] Created TCP listener on port 56628
DEBUG:asyncssh:[conn=0] Assigning dynamic port 56628
INFO:asyncssh:[conn=1] Opening forwarded TCP connection to port 56628
INFO:asyncssh:[conn=1]   Client address: ::1, port 56629
DEBUG:asyncssh:[conn=1, chan=0] Set write buffer limits: low-water=16384, high-water=65536
DEBUG:asyncssh:[conn=1, chan=0]   Initial recv window 2097152, packet size 32768
INFO:asyncssh:[conn=0] Accepted forwarded TCP connection on port 56628
INFO:asyncssh:[conn=0]   Client address: ::1, port 56629
DEBUG:asyncssh:[conn=0, chan=0]   Initial send window 2097152, packet size 32768
INFO:asyncssh:[conn=0]   Forwarding TCP connection to port 56626
DEBUG:asyncssh:[conn=0, chan=0]   Initial recv window 2097152, packet size 32768
DEBUG:asyncssh:[conn=0, chan=0] Reading from channel resumed
DEBUG:asyncssh:[conn=1, chan=0]   Initial send window 2097152, packet size 32768
DEBUG:asyncssh:[conn=1, chan=0] Reading from channel resumed
DEBUG:asyncssh:[conn=1, chan=0] Sending 11 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 11 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Sending 11 data bytes
DEBUG:asyncssh:[conn=1, chan=0] Received 11 data bytes
INFO:asyncssh:[conn=0] Creating remote TCP forwarder from port 56628 to port 56626
INFO:asyncssh:[conn=0] Creating remote TCP listener on port 56628
INFO:asyncssh:[conn=1] Creating local TCP forwarder on port 56628
DEBUG:asyncssh:[conn=1, chan=0] Sending EOF
DEBUG:asyncssh:[conn=1] Failed to create local TCP listener: [Errno 48] error while attempting to bind on address ('::', 56628, 0, 0): Address already in use
DEBUG:asyncssh:[conn=1] Failed to create TCP listener
DEBUG:asyncssh:[conn=0] Failed to create remote TCP listener
DEBUG:asyncssh:[conn=0, chan=0] Sending EOF
INFO:asyncssh:[conn=0, chan=0] Closing channel
INFO:asyncssh:[conn=1] Closed TCP listener on port 56628
DEBUG:asyncssh:[conn=1, chan=0] Received EOF
INFO:asyncssh:[conn=1, chan=0] Received channel close
INFO:asyncssh:[conn=1, chan=0] Closing channel
INFO:asyncssh:[conn=1, chan=0] Channel closed
INFO:asyncssh:[conn=0] Closed remote TCP listener on port 56628
INFO:asyncssh:[conn=0] Creating remote TCP forwarder from port 56628 to port 56626
INFO:asyncssh:[conn=0] Creating remote TCP listener on port 56628
INFO:asyncssh:[conn=1] Creating local TCP forwarder on port 56628
INFO:asyncssh:[conn=1] Created TCP listener on port 56628
INFO:asyncssh:[conn=1] Closed TCP listener on port 56628
INFO:asyncssh:[conn=0] Closed remote TCP listener on port 56628
```

Note how the second attempt to open a listener fails because the first listener is still open, but the third attempt succeeds, since it is after the close of the first listener.

I'm going to try to put together a similar test against OpenSSH as a server to see if I can find any problem there, but sending the hostname string without prefixing it with a length definitely goes against the RFC.

ronf:
Here's some code I used to test this against an OpenSSH server. It works fine for me with the currently released AsyncSSH:

```python
import asyncio, asyncssh, logging, sys

async def run_client():
    logging.basicConfig()
    asyncssh.set_log_level('DEBUG')
    asyncssh.set_debug_level(1)

    async with asyncssh.connect('localhost') as conn:
        listener = await conn.forward_remote_port('', 8080, 'localhost', 80)

        listener2 = await conn.forward_remote_port('', 8080, 'localhost', 80)
        assert(listener2 == None)

        listener.close()
        await listener.wait_closed()

        await asyncio.sleep(0.2)

        listener = await conn.forward_remote_port('', 8080, 'localhost', 80)

        assert(listener != None)

        listener.close()
        await listener.wait_closed()

try:
```

Here's a snippet of the log output I got:

```
INFO:asyncssh:[conn=0] Creating remote TCP forwarder from port 8080 to localhost, port 80
INFO:asyncssh:[conn=0] Creating remote TCP listener on port 8080
DEBUG:asyncssh:[conn=0] Received unknown global request: hostkeys-00@openssh.com
INFO:asyncssh:[conn=0] Creating remote TCP forwarder from port 8080 to localhost, port 80
INFO:asyncssh:[conn=0] Creating remote TCP listener on port 8080
DEBUG:asyncssh:[conn=0] Failed to create remote TCP listener
INFO:asyncssh:[conn=0] Closed remote TCP listener on port 8080
INFO:asyncssh:[conn=0] Creating remote TCP forwarder from port 8080 to localhost, port 80
INFO:asyncssh:[conn=0] Creating remote TCP listener on port 8080
INFO:asyncssh:[conn=0] Closed remote TCP listener on port 8080
```

As with my other test, the second attempt to create a listener fails because the first one is still open, but the third attempt succeeds, since it is after the close call, showing that the OpenSSH server does respond correctly to this request.

You can also see the hostkeys-00@openssh.com message in here, indicating that this was in fact talking to an OpenSSH server and not AsyncSSH as in the previous test.

Itayazolay:
Thank you for your response. 
This is weird, because I have similar code that does exactly that except mine fails.
I will try to check my environment and test it again, and close the issue if it works.
Thank you

Itayazolay:
Hi, 
Sorry for leaving this issue open for a long time.
I've used your example code, and it really did work for me on OpenSSH Server.
I've also tested it with a dropbear server - and it didn't work. 
It is probably a problem in dropbear implementation. 
Thank you again for your fast and detailed response, and sorry for rushing to blame asyncssh implementation 😅

ronf:
No problem!

I did a little digging into dropbear today to see if I could find a problem there. There are no issues with parsing the messages, but it turns out that there are actually multiple issues that can prevent dropbear from properly closing a remote TCP listener, and unfortunately it doesn't provide any error indication when these things happen.

The first issue is just a bug. The svr_cancelremotetcp() function calls get_listener() to try and match on the listener to close, passing in a function matchtcp() to do the matching. In addition to address and port, that function expects to match on "chantype", but the "chantype" field of the tcpinfo value passed in to match against is never set to anything. So, the match fails. It should be set to &svr_chan_tcpremote, just like it is in svr_remotetcpreq().

Unfortunately, even after fixing this, there's another problem. If the "-a" argument is not passed to dropbear on startup or if the listening request uses a hostname of "localhost", the listener that's created ends up having its listening address set to NULL. This restricts inbound connections to only be from localhost, which is just fine. However, when the matching is done later in svr_cancelremotetcp() to close the listener, there's no equivalent code to set the listener to NULL in this case, so the match fails.

There's actually a bigger problem in this second case, in fact. The matchtcp() function does a strcmp() of the listening addresses without checking if they are NULL first, causing a segmentation fault when it dereferences the NULL pointer if "-a" is not provided at startup or the listen host passed in is "localhost".

By fixing the bug of not filling in the chantype, passing in the "-a" argument to dropbear, and avoiding the use of "localhost" as the listen host, I've been able to get the test discussed here to work without any changes to AsyncSSH. However, without patching dropbear, I don't think there's any way you're going to get it to properly handle the cancel-tcpip-forward request and close the remote listener.

PR #158, opened 2018-07-21
Do not use root logger when installed as dependency

### LINKED ISSUE #156: asyncssh Library Logs Using Root Logger When Installed From Python Egg
Meaning that you can not set silence asyncssh's logging

The following code demonstrates the issue:

```
import logging
import asyncssh.logging

root_logger = logging.getLogger()

def print_root_logger_level():
    print(logging.getLevelName(root_logger.level))

print_root_logger_level()

asyncssh.logging.set_log_level(logging.DEBUG)

print_root_logger_level()

print(asyncssh.logger)
```

The output is
```
WARNING
DEBUG
<_SSHLogger root (DEBUG)>
```

So the default level of root logger is WARNING and after setting log level using asyncssh.logging.set_log_level to DEBUG also the root logger level is changed.

Also all the messages logged by asyncssh seems to be logged through the root logger.

Relevan version and environment:
asyncssh 1.13.2
Python 3.7
Archlinux

I wasn't able to reproduce this issue when running with asyncssh from Git, i.e. with
```
python -m venv venv
venv/bin/python setup.py develop
```

I then packaged asyncssh into wheel myself with `python setup.py bdist_wheel`. When installed from the wheel the above test does what it is supposed to do so the fix would be just to distribute asyncssh as wheel instead of egg.

The `asyncssh.logging` uses `__package__` to define the used logger and I'm guessing that when the code is run from Python egg the `__package__` is an empty string. I did not find confirmation for this with a quick search though

jannero:
This fixes #156 without needing to change release package format.

### DIFF
--- a/asyncssh/logging.py
+++ b/asyncssh/logging.py
@@ -20,7 +20,11 @@ class _SSHLogger(logging.LoggerAdapter):
     """Adapter to add context to AsyncSSH log messages"""
 
     _debug_level = 1
-    _pkg_logger = logging.getLogger(__package__)
+    # If __package__ is used, root logger will be used if asyncssh is installed
+    # as dependency because in that case the distributed egg file is placed
+    # into site-packages as is (without extracting the files).
+    # See https://github.com/ronf/asyncssh/issues/156
+    _pkg_logger = logging.getLogger("asyncssh")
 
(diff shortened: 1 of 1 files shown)


ronf:
Thanks! I just checked this fix into the develop branch.

PR #160, opened 2018-07-29
Guard some imports and tests for Fedora 27

gsauthof:
such that `python3 setup.py test` executes successfully.

Background: I'm packaging asyncssh for Fedora. And in that process I noticed some test cases failing.

### DIFF
--- a/tests/sspi_stub.py
+++ b/tests/sspi_stub.py
@@ -12,8 +12,12 @@
 
 """Stub SSPI module for unit tests"""
 
-from asyncssh.gss_win32 import ASC_RET_INTEGRITY, ISC_RET_INTEGRITY
-from asyncssh.gss_win32 import SECPKG_ATTR_NATIVE_NAMES, SSPIError
+import sys
+# on non-windows systems, work around setup.py test discovery
+# importing everything
+if sys.platform == 'win32':
+    from asyncssh.gss_win32 import ASC_RET_INTEGRITY, ISC_RET_INTEGRITY
+    from asyncssh.gss_win32 import SECPKG_ATTR_NATIVE_NAMES, SSPIError
 
 from .gss_stub import step
 
--- a/tests/test_public_key.py
+++ b/tests/test_public_key.py
@@ -26,6 +26,7 @@ from pathlib import Path
 import shutil
 import subprocess
 import sys
+import unittest
 
 import asyncssh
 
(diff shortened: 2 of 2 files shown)


ronf:
Could you provide me the output of "openssl ecparam -list_curves" on your system? While a guard like that is still a good idea, I might be able to pick a different curve that would work on a wider variety of OpenSSL versions.

Regarding the sspi_stub issue, how are you invoking the tests? I don't see that error here when I invoke "tox", as the import of sspi_stub is already guarded in tests/util.py by a check against "sys.platform == 'win32'". It sounds in your case like that module is being imported after being found by discovery, but since that module doesn't have "test" in its name, I'm surprised by that.

ronf:
Huh - that's a really short list. I get 80-100 lines of output on the few different versions of OpenSSL I have installed on the Mac, and pretty much the same on Ubuntu 16.04. Any idea why it's so restricted on Fedora/CentOS?

It looks like "prime256v1" is an alternate name for "secp256r1", which is one of the three standard curves that SSH supports along with secp384r1 and secp512r1. Since I've now added support for secp256k1 as well, though, there's nothing left which is supported by the CentOS 7 OpenSSL that I could use to perform this test. I could switch from secp112r1 to secp224r1 to allow the test to work on Fedora, though.

I tried running "setup.py test" here and was able to reproduce the failure you mentioned, so I'll look into that further. The easiest thing to do is to probably replicate my existing test of sys.platform in that module directly (similar to your fix, but following the standard I use elsewhere for testing for win32).

I also noticed the AppVeyor and TravisCI builds both failed on your patch. It didn't get the expected KeyImportError. It looks like you left out the "run" command to have OpenSSL generate the unsupported key.

gsauthof:
Yes, the Fedora openssl package, there are some patches that disable some curves. Apparently, this is motivated by some [complicated legal situation](https://bugzilla.redhat.com/show_bug.cgi?id=319901#c36) (involving patents or something like that; 2013 quote from the changelog 'only ECC NIST Suite B curves support'). Another motivation might be that some curves are considered insecure - but haven't found any fedora documentation about this point.

Yes, the `run()` part was lost during editing - thus, I've rebase this pull-request to include it again. Also, switched the windows guard to the style you use in other files. The CI hooks are now active again.

ronf:
Thanks very much for the pointer to the history there - I wasn't aware of any of that.

Also, thanks for updating your pull request. I've merged it with some minor edits to address pylint issues into the "develop" branch, and I'll merge it back into "master" when I cut the next release.

PR #183, opened 2019-01-17
Use the default known_hosts file only if it exists

kvalev:
Currently the code makes the assumption that a default `.ssh/known_hosts` exists, which is not always case (especially when using the official Python docker containers).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 MANIFEST
 __pycache__/
 *.py[cod]
+.tox
 
 asyncssh.egg-info
 build/
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -2230,8 +2230,10 @@ class SSHClientConnection(SSHConnection):
             self._trusted_ca_keys = None
         else:
             if not self._known_hosts:
-                self._known_hosts = os.path.join(os.path.expanduser('~'),
-                                                 '.ssh', 'known_hosts')
+                default_known_hosts = os.path.join(os.path.expanduser('~'),
+                                                   '.ssh', 'known_hosts')
+                if os.path.isfile(default_known_hosts):
+                    self._known_hosts = default_known_hosts
 
             port = self._port if self._port != _DEFAULT_PORT else None
 


ronf:
Thank you for the pull request. I don't think this will work as written, though. When I try it with no ~/.ssh/known_hosts file, I get the following error:

```
  File "./simple_client.py", line 40, in <module>
    loop.run_until_complete(run_client(loop))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/asyncio/base_events.py", line 473, in run_until_complete
    return future.result()
  File "./simple_client.py", line 35, in run_client
    agent_forwarding=True) as conn:
  File "/Users/ronf/src/asyncssh/tmp/asyncssh/misc.py", line 186, in __aenter__
    self._result = yield from self._coro
  File "/Users/ronf/src/asyncssh/tmp/asyncssh/connection.py", line 5448, in connect
    conn, _ = yield from create_connection(None, host, port, **kwargs)
  File "/Users/ronf/src/asyncssh/tmp/asyncssh/connection.py", line 5104, in create_connection
    yield from auth_waiter
  File "/Users/ronf/src/asyncssh/tmp/asyncssh/connection.py", line 587, in connection_made
    self._connection_made()
  File "/Users/ronf/src/asyncssh/tmp/asyncssh/connection.py", line 2242, in _connection_made
    self._peer_addr, port)
  File "/Users/ronf/src/asyncssh/tmp/asyncssh/connection.py", line 473, in _match_known_hosts
    match_known_hosts(known_hosts, host, addr, port)
ValueError: not enough values to unpack (expected 7, got 0)
```

This is expected, as leaving known_hosts set to () instead of a proper 7-tuple returned by match_known_hosts isn't right. However, I think something like the following might work instead:

```patch
index 51a2fa3..0cdfb9d 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -2233,8 +2233,13 @@ class SSHClientConnection(SSHConnection):
             self._trusted_ca_keys = None
         else:
             if not self._known_hosts:
-                self._known_hosts = os.path.join(os.path.expanduser('~'),
-                                                 '.ssh', 'known_hosts')
+                default_known_hosts = os.path.join(os.path.expanduser('~'),
+                                                   '.ssh', 'known_hosts')
+
+                if os.path.isfile(default_known_hosts):
+                    self._known_hosts = default_known_hosts
+                else:
+                    self._known_hosts = b''
 
             port = self._port if self._port != _DEFAULT_PORT else None
 
```

Setting it to an empty byte string will trigger the import_known_hosts() code path on that empty string, resulting in no hosts being trusted but the proper set of values being returned.

What's your use case? Generally speaking, this will just result in a "Host key is not trusted" error if you don't have a known_hosts file and don't specify known hosts explicitly or set it to `None` to disable this checking. Is your intent to use the newly-added callback-based host key validation?

I guess the other possibility is that you've specified trusted X.509 CA certs via x509_trusted_certs. This could be done before by explicitly specifying known_hosts='' as an argument along with that, but this change would make that unnecessary.

ronf:
This fix is now available in the "develop" branch as commit f23209e. It'll be included in the next AsyncSSH release. It's similar to what is show above, but I also added a check that the file is readable before using it, and added unit tests.

Thanks again for submitting this!

kvalev:
You are correct - the error resulted due to an incorrect usage of the API (basically forgot to explicitly set the known hosts to `None`), but thought that such a check would be helpful anyway.
Thanks for fixing and merging the issue.

ronf:
Yeah - I think this could be helpful in cases where GSS or X.509 is used exclusively and known hosts entries are never created. Thanks for suggesting it.

One thing: Be careful setting known_hosts to None, as that means to disable known_hosts checking completely. This potentially opens the client up to a man-in-the-middle attack.

PR #194, opened 2019-03-06
Test ed25519 and curve448 only if is available

mimi1vx:
ed25519 now uses cryptography so test based on nacl presence is obsolete

curve448 - disable tests if this curve isn't available


ronf:
A variation of this is now available as commit 2b86e6e in the develop branch. Thanks very much!

This will be included in the next release.

PR #199, opened 2019-03-10
Optimize SFTP glob matching

termim:
This commit reduces number of calls to listdir() if glob pattern
contains directory names without glob characters. This could improve
performance significantly if the network latency is high and/or the
intermediate directories have big number of entries.

For example before this commit the pattern "/usr/share/doc/qt5/global/*"
resulted in six listdir() calls but with this commit only one call is
needed. And the total number of returned entries reduced from more
than 1.5K to just couple of dozen in my case.

An additional bonus is that with this commit the case when mount point
of the auto-mounted directory is used without glob characters in the
pattern then it is processed correctly.

### DIFF
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -147,44 +147,73 @@ def _setstat(path, attrs):
         os.utime(path, times=(attrs.atime, attrs.mtime))
 
 
+def _split_path_by_globs(pattern):
+    """Split path grouping parts without glob pattern"""
+
+    basedir, patlist, plain = None, [], []
+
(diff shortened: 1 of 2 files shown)


ronf:
This is now merged into the "develop" branch. I refactored the test code to cut down on the duplication and make it table driven, added a test case to get back full code coverage, and did some other minor cosmetic cleanup, but the patch is otherwise unchanged.

Thanks very much for submitting this! It'll be in included in the next release.

termim:
Thank you, I really enjoy using asyncssh!

PR #200, opened 2019-03-15
Allow PurePath for the single file source argument in SFTP get, put and copy

termim:
PurePath was missing in the check if source is a single file.


ronf:
Looks good - this is now in the "develop" branch as commit 4b7e61a. Thanks very much!

PR #201, opened 2019-03-20
Allow PurePath for the source and target arguments in scp

termim:
Using PurePath is the only way to use paths with a drive letter
on Windows.


ronf:
Looks good - this is now commit 17983b3 in "develop". Thanks again!

PR #239, opened 2019-11-26
channel: Make check for None explicit

PromyLOPh:
I’m trying to write a [connector](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py) for aiohttp, which SSH’s into a machine and then connects to a local UNIX domain socket on that host. While debugging why my BaseProtocol’s .data_received method was never called I noticed that SSHChannel._deliver_data implicitly checks self._session’s truth value, but not whether it is None. Since the BaseProtocol implements \_\_len\_\_, which can be zero, no data is forwarded to the callback.

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -185,7 +185,7 @@ class SSHChannel(SSHPacketHandler):
 
             self._request_waiters = []
 
-        if self._session:
+        if self._session is not None:
             self._session.connection_lost(exc)
             self._session = None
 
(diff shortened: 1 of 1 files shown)


ronf:
I'm not seeing an implementation of __len__ in asyncio.BaseProtocol. Is this something specific to the subclass of it you are creating? I don't mind making this check more explicit, but what would it mean to call len() on a Protocol object? Given the set of operations allowed on a Protocol, I don't really see a len() operation as making sense.

PromyLOPh:
I wasn’t specific enough, sorry. aiohttp’s [ResponseHandler](https://github.com/aio-libs/aiohttp/blob/dd85639f0e1855d8921c57db8643b28ffe3f6b25/aiohttp/client_proto.py#L17), which inherits from BaseProtocol, is also a [DataQueue](https://github.com/aio-libs/aiohttp/blob/dd85639f0e1855d8921c57db8643b28ffe3f6b25/aiohttp/streams.py#L536). That one defines a \_\_len\_\_ method.

ronf:
Thanks for the explanation, and the proposed fix.

This change is now available as commit ac96f44 in the "develop" branch, and will be included in the next release.

PR #242, opened 2019-12-19
typo in docstring

termim:


ronf:
Thanks for catching this. This is now fixed in the "develop" branch as commit 7b73a4e, and will be included in the next release.

PR #243, opened 2019-12-30
Update pywin32 dependency

nuno-andre:
`pypiwin32` is an outdated package [created by the same author](https://pypi.org/user/mhammond/) to provide wheels. The only maintained package is currently [`pywin32`](https://pypi.org/project/pywin32/#history).


ronf:
Thanks for pointing this out! I have updated this and some other references to the pypiwin32 package in the "develop" branch in commit d771d64, and this will be rolled into the next release.

PR #261, opened 2020-02-20
Issue-259: Check that peername is not None

### LINKED ISSUE #259: ConnectionResetError: [Errno 104] Connection reset by peer
I get the above error when running the following code:

```python
await asyncio.wait_for(
    asyncssh.connect(
        hostname,
        password=...,
        username=...,
        known_hosts=None,
        client_keys=[....],
    ),
    timeout=5,
)
```
The traceback shows that the problem originated with peername being None
```
Exception in callback SSHConnection.connection_made(<_SelectorSoc...e, bufsize=0>>)
handle: <Handle SSHConnection.connection_made(<_SelectorSoc...e, bufsize=0>>)>
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File ".../lib/python3.8/site-packages/asyncssh/connection.py", line 692, in connection_made
    self._peer_addr, self._peer_port = peername[:2]
TypeError: 'NoneType' object is not subscriptable
```

I found [this](https://github.com/ronf/asyncssh/issues/16#issuecomment-106666655) issue where you explained that peername can be None and that it's best to check if it's None before attempting to split it.
So is it as a simple as checking if peername is None and if it is skipping over that line? If so, I'd be happy to make a PR with a fix.

d1618033:
fixes #259

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -689,7 +689,8 @@ class SSHConnection(SSHPacketHandler):
         self._local_addr, self._local_port = sockname[:2]
 
         peername = transport.get_extra_info('peername')
-        self._peer_addr, self._peer_port = peername[:2]
+        if peername is not None:
+            self._peer_addr, self._peer_port = peername[:2]
 
         self._owner = self._protocol_factory()
         self._protocol_factory = None


ronf:
Thanks - a variation of this fix is available in commit a7e5bfb.

PR #269, opened 2020-03-14
handle `SSHServer.server_requested` Bool returns

tommyvn:
`SSHServer.server_requested` is documented to allow a coroutine return, but prior to this commit a return of False from the coroutine bypassed the code that handles the synchronous version.

Further, adding `True` to the async flow proved trivial and so that is done too.

This is my first PR here so I'm adding that, if accepted, I'm happy to contribute this change under the Eclipse Public License v2.0 and that I'm leaving how/if you list me in the contributors list entirely up to you.

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -4343,27 +4343,27 @@ class SSHServerConnection(SSHConnection):
 
         result = self._owner.server_requested(listen_host, listen_port)
 
-        if not result:
-            self.logger.info('Request for TCP listener on %s denied by '
-                             'application', (listen_host, listen_port))
-
-            self._report_global_response(False)
-            return
-
-        if result is True:
-            result = self.forward_local_port(listen_host, listen_port,
-                                             listen_host, listen_port)
-
         self.create_task(self._finish_port_forward(result, listen_host,
                                                    listen_port))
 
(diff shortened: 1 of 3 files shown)


ronf:
Thanks very much for your submission, Tom!

I've gone ahead and checked your proposed changes with a few minor adjustments into the "develop" branch as commit 17fd581. I moved the position of the try..except block that was looking for OSError, edited the documentation slightly, and fixed some issues with parentheses and line length, but this should preserve the functionality you added to properly handle True and False being returned from an async version of the server_requested() method.

Also, thanks very much for adding unit tests for the async version of server_requested() as part of your change!

If you get a chance, could you give the "develop" branch a try to confirm everything is working as you expect?

tommyvn:
All good, and thank you for the other thousands of lines of code you've created! I'm just tinkering with it at the moment but it's a hella useful framework.

`debug1: remote forward failure for: listen localhost.localdomain:80, connect localhost:8001` so says my ssh client against that commit id, which means it's working on your check in to develop, so consider this confirmation it's now resolved. I'll close this PR, thanks!

ronf:
Thanks very much for contributing this, and confirming the revised change. This will be included in the next release!

PR #277, opened 2020-04-27
Add support for ip_switch file xfer code

erichoxworth-renaissance:
asyncssh (sftp) was failing for me when communicating with ip_switch SFTP server.  The failure occurred during the SFTPAttrs.decode method on line 855 when a flag value of 0x0000003d was received from the ip_switch SFTP server. afaict code 0x0000003d (dec 61) is a valid non-error code (https://support.microfocus.com/kb/doc.php?id=7021956).  This PR changes the check mask to allow 0x0000003d as a valid response code.

Note that other libs and clients handle this response correctly.

### DIFF
--- a/asyncssh/constants.py
+++ b/asyncssh/constants.py
@@ -143,7 +143,7 @@ FILEXFER_ATTR_UIDGID                = 0x00000002
 FILEXFER_ATTR_PERMISSIONS           = 0x00000004
 FILEXFER_ATTR_ACMODTIME             = 0x00000008
 FILEXFER_ATTR_EXTENDED              = 0x80000000
-FILEXFER_ATTR_UNDEFINED             = 0x7ffffff0
+FILEXFER_ATTR_UNDEFINED             = 0x7fffffc0
 
 # OpenSSH statvfs attribute flags
 FXE_STATVFS_ST_RDONLY               = 0x1


ronf:
I'd be happy to try and make AsyncSSH work better with ip_switch, but I need more information.

Your comment talks about error codes, but your proposed code change is actually to FILEXFER_ATTR_UNDEFINED, which is a bitmask for attributes and has nothing to do with error codes. From your change, it seems like you are attempt to allow additional attributes with bit values of 0x10 and 0x20, but it doesn't make sense to change that mask without knowing what type of attributes those bits represent and what their encoding is on the wire. Otherwise, even though the initial check of the bit mask may not fail, it's very likely it will fail later when AsyncSSH finds there's attribute data left in the packet which it hasn't decoded.

While later versions of the SFTP spec define additional attributes in this bit mask, AsyncSSH supports version 3 of SFTP, defined at https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02#section-5 and it only defines 0x01-0x08 and 0x80000000 (for extended attributes). Does the ip_switch implementation negotiate SFTP version 3, and if so does it still send attributes beyond what is defined in the spec? If so, is there any documentation on their non-standard extension?

If you look at the decode function more closely, you'll see that each bit which is defined causes specific attribute data to be decoded. So, if you take bits out of the UNDEFINED mask, you really need to add lines for each of those bits to tell it how to unpack the corresponding attributes. If you don't do that, calls to packet.check_end() such as in SFTPClientHandler._process_attrs line 1258 will probably still cause a failure.

Here's what the SFTP spec has to say about extending the bitmask:

     Additional fields can be added to the attributes by either defining
     additional bits to the flags field to indicate their presence, or by
     defining extended attributes for them.  The extended attributes
     mechanism is recommended for most purposes; additional flags bits
     should only be defined by an IETF standards action that also
     increments the protocol version number.  The use of such new fields
     MUST be negotiated by the version number in the protocol exchange.
     It is a protocol error if a packet with unsupported protocol bits is
     received.

ronf:
Closing this for now, due to lack of information.

PR #310, opened 2020-09-03
Fixes #309 Allows _cleanup() to close listeners on an SSHClientConnection

### LINKED ISSUE #309: SSHClientConnection not closing listeners on error when using forward_local_port()
I am using the following example from the documentation

```
import asyncio, asyncssh, sys

async def run_client():
    async with asyncssh.connect('my.host.com') as conn:
        listener = await conn.forward_local_port('', 9292, 'other.host.com', 10040)
        await listener.wait_closed()
try:
```

When I kill the ssh connection on the remote server with `ss -K ..` the logs report the connection as closed, but I get no exceptions raised and wait_closed() does not return.

```
...
DEBUG:asyncssh:[conn=0] Got keepalive response
DEBUG:asyncssh:[conn=0] Sending keepalive request
DEBUG:asyncssh:[conn=0] Got keepalive response
INFO:asyncssh:[conn=0] Connection failure: [WinError 64] The specified network name is no longer available
```
*Edit*

There doesn't seem to be any way to pass in an error handler function, but manually adding one, and using this to close the listener on error does fix the problem. 

```
    async with asyncssh.connect('my.host.com') as conn:

        listener = await conn.forward_local_port('', 9292, 'other.host.com', 10040)

        def close_listener(*args,**kwargs):
            print("close listener")
            listener.close()
        conn._error_handler = close_listener

        await listener.wait_closed()
```
It does seem like the listener should be automatically closed when the connection is closed, but perhaps not, and the current behaviour is as intended?

*Edit*

It seems the proper way to get notifications on connection status is use `SSHClientConnectionOptions(client_factory=...)`, sublcassing `SSHClient` overriding `connection_made` storing the `SSHClientConnection` connection instance that is passed and overriding `connection_lost` to handle the error.

However after playing around it seems like the listener is not added to the `self._local_listeners` dictionary on the `SSHClientConnection` when using `forward_local_port()`. The dictionary is iterated in `_cleanup` [here](https://github.com/ronf/asyncssh/blob/master/asyncssh/connection.py#L575-L576), where the listeners are all closed.

adding `self._local_listeners[listen_host, listen_port] = listener` [here](https://github.com/ronf/asyncssh/blob/master/asyncssh/connection.py#L2436) solves the problem and it now exits properly on connection failure.

The same problem possibly exists in `forward_local_path` although I have not tested it.

Fixed by [#310](https://github.com/ronf/asyncssh/pull/310)

rmawatson:
Allows _cleanup to close listeners on an SSHClientConnection
[Issue 309](https://github.com/ronf/asyncssh/issues/309)

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -2433,10 +2433,12 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
                                                          tunnel_connection,
                                                          listen_host,
                                                          listen_port)
-
+            
+            self._local_listeners[listen_host, listen_port] = listener
+                
             if dest_port == 0:
                 dest_port = listener.get_port()
-
+            
             return listener
         except OSError as exc:
             self.logger.debug1('Failed to create local TCP listener: %s', exc)


ronf:
Thanks for your proposed fix! This is now available in commit 869cea0.

PR #311, opened 2020-09-08
Allow connection timeouts to work as expected

ashafer01:
While testing an asyncssh script against a known broken host that times out with openssh, I found that the timeout never occurred (following `asyncio.wait_for` advice from the documentation) and this exception was thrown immediately:

```
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 910, in data_received
    while self._inpbuf and self._recv_handler():
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 1143, in _recv_packet
    processed = handler.process_packet(pkttype, seq, packet)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/packet.py", line 215, in process_packet
    self._packet_handlers[pkttype](self, pkttype, pktid, packet)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 551, in _process_eof
    self._flush_recv_buf()
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 305, in _flush_recv_buf
    if (not self._session.eof_received() and
AttributeError: 'NoneType' object has no attribute 'eof_received'
```

This patch prevents the exception and allows the timeout to occur.

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -302,7 +302,8 @@ class SSHChannel(SSHPacketHandler):
             if self._recv_state == 'eof_pending':
                 self._recv_state = 'eof'
 
-                if (not self._session.eof_received() and
+                if (self._session and
+                        not self._session.eof_received() and
                         self._send_state == 'open'):
                     self.write_eof()
             elif self._recv_state == 'close_pending':


ronf:
I'm happy to put in the extra safeguard, but I'd like to understand a bit more about how this might have happened. The code in question is already guarded by a check against the recv_state being "eof_pending", which suggests that the channel was opened  and a session object was created for it.

If the receive state is 'eof_pending' when the code in question runs, it was probably triggered by the remote system sending an EOF indication, but the _session value for the channel should only be set back to `None` in the _cleanup() function, so that would suggest that incoming packets were processed after the channel was cleaned up, which shouldn't really be possible. The _cleanip() function does a remove_channel() of itself from the SSHConnection object, and without that entry any message received for that channel would trigger a ProtocolError.

Do you have a reliable way to reproduce this error?

ashafer01:
I will see about getting this reproducible. One confounding factor may be the use of tunnel connections / jump hosts. When this occurred, it also caused the tunnel connection to close, and subsequent connections that attempted to use the same tunnel connection object failed. When it's allowed to time out, the tunnel stays alive and works for subsequent connections.

ashafer01:
Now realizing that this output is probably relevant, too:
```
future: <Future finished exception=AttributeError("'NoneType' object has no attribute 'eof_received'") created at /usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py:422>
source_traceback: Object created at (most recent call last):
  File "macos_image_manager.py", line 679, in <module>
    asyncio.run(main(), debug=True)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 1851, in _run_once
    handle._run()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/tasks.py", line 684, in _wrap_awaitable
    return (yield from awaitable.__await__())
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 6414, in connect
    return await _connect(options.host, options.port, loop, options.tunnel,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 209, in _connect
    _, conn = await tunnel.create_connection(conn_factory, host, port)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 3510, in create_connection
    session = await chan.connect(session_factory, remote_host, remote_port,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 1868, in connect
    return (await self._open_tcp(session_factory, b'direct-tcpip',
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 1860, in _open_tcp
    return (await super()._open_forward(session_factory, chantype,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 1841, in _open_forward
    self._session = session_factory()
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 6405, in conn_factory
    return SSHClientConnection(loop, options, wait='auth')
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 2538, in __init__
    super().__init__(loop, options, acceptor, error_handler,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 418, in __init__
    self._waiter = loop.create_future()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 422, in create_future
    return futures.Future(loop=self)
```

ronf:
The traceback looks pretty normal for a call to asyncssh.connect(), and it is expected that an exception which is caught while waiting for the connection to open could eventually be passed into the _waiter future object that the connect() call eventually blocks on. As you noted, this connect appears to be opening a tunnel, as it is actually using the create_connection() call on the SSHClientConnection object, rather than directly opening a TCP connection when making the connection.

Out of curiosity, if you pass in a login_timeout instead of using wait_for(), do you still see the same problem?

ashafer01:
Yes, passing `login_timeout`to `asyncssh.connect` yields the same un-awaited exception (at least approximately, have not compared every line)

```
Future exception was never retrieved
future: <Future finished exception=AttributeError("'NoneType' object has no attribute 'eof_received'") created at /usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py:422>
source_traceback: Object created at (most recent call last):
  File "macos_image_manager.py", line 694, in <module>
    asyncio.run(main(), debug=True)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 1851, in _run_once
    handle._run()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "macos_image_manager.py", line 343, in host_prep_worker
    host_plan = await self.host_prep_task(hostname, gateway)
  File "macos_image_manager.py", line 356, in host_prep_task
    async with self.create_host(hostname, gateway) as host:
  File "macos_image_manager.py", line 443, in __aenter__
    await self._connect()
  File "macos_image_manager.py", line 435, in _connect
    self._conn = await asyncssh_connect(self.hostname,
  File "macos_image_manager.py", line 661, in asyncssh_connect
    return await asyncssh.connect(hostname, **kwds)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 6414, in connect
    return await _connect(options.host, options.port, loop, options.tunnel,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 209, in _connect
    _, conn = await tunnel.create_connection(conn_factory, host, port)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 3510, in create_connection
    session = await chan.connect(session_factory, remote_host, remote_port,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 1868, in connect
    return (await self._open_tcp(session_factory, b'direct-tcpip',
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 1860, in _open_tcp
    return (await super()._open_forward(session_factory, chantype,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 1841, in _open_forward
    self._session = session_factory()
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 6405, in conn_factory
    return SSHClientConnection(loop, options, wait='auth')
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 2538, in __init__
    super().__init__(loop, options, acceptor, error_handler,
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 418, in __init__
    self._waiter = loop.create_future()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 422, in create_future
    return futures.Future(loop=self)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 910, in data_received
    while self._inpbuf and self._recv_handler():
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 1143, in _recv_packet
    processed = handler.process_packet(pkttype, seq, packet)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/packet.py", line 215, in process_packet
    self._packet_handlers[pkttype](self, pkttype, pktid, packet)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 551, in _process_eof
    self._flush_recv_buf()
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 305, in _flush_recv_buf
    if (not self._session.eof_received() and
AttributeError: 'NoneType' object has no attribute 'eof_received'
```

The PR also appears to fix this case.

ronf:
Ok, thanks for checking on that. One other thought: If you don't set any timer with wait_for and disable the login timeout by setting it to None but you try to catch the AttributeError on the connect() call, does that actually get retrieved? That's not the right thing to have to do -- I'm just curious if the AsyncSSH code is at least properly propagating that error out to the application code. From the traceback here, I think it probably is, but your code isn't trying to catch it and so your application task calling connect() ends up being killed.

I'm still looking through the code to see how _session could end up being set to None after the connection is established. When you say the SSH server is "broken", what does that mean exactly? Have you tried enabling debugging to see how far it gets in the SSH connect? You can do that with code like:

```python
    import logging
    logging.basicConfig()
    asyncssh.set_log_level('DEBUG')
    asyncssh.set_debug_level(2)
```

ashafer01:
Confirmed that a `try/except AttributeError` wrapped tightly around the `asyncssh.connect` call **does** receive the AttributeError, and the "Future exception was never retrieved" still occurs with `_waiter` and the same exception.

This appears to be all of the relevant debug logging:
```
INFO:asyncssh:[conn=1] Opening SSH connection to 172.17.4.15, port 22 via SSH tunnel
INFO:asyncssh:[conn=1] Opening direct TCP connection to 172.17.4.15, port 22
```

```
INFO:asyncssh:[conn=9] Connection to 172.17.4.15, port 22 succeeded
INFO:asyncssh:[conn=9]   Local address: 192.168.2.95, port 59723
INFO:asyncssh:[conn=9] Connection lost
INFO:asyncssh:[conn=9] Aborting connection
```

We're not sure exactly what is wrong with the server, other than we can't SSH to it at the moment. We've been allowing it to stay broken for now during this development effort, as it makes such a good test case 🙂

ronf:
Thanks for confirming that - I'm glad to hear that an unexpected error like this is at least being property propagated back, if the application code is doing exception handling.

Regarding the debug logs, it seems like there is something listening for TCP connections on port 22 on the "broken" system, but it closes the socket immediately after the TCP handshake completes, without even sending back an SSH version string. I imagine if you were to telnet (or 'nc') to that system on port 22, you might be able to confirm that behavior.

Based on this, it seems like the error we're seeing actually returns to the channel being opened on the tunnel connection. Looking at the debug messages for the tunnel connection, I see the following when the remote system closes the forwarded TCP connection:

```
DEBUG:asyncssh:[conn=0, chan=0] Sending 24 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Reading from channel started
INFO:asyncssh:[conn=0, chan=0] Aborting channel
```

The 24 data bytes is the SSH version string we send, but all we get back is an EOF, and that triggers a call to _flush_recv_buf() where you saw the problem. So, apparently, there's a race between a MSG_CHANNEL_EOF being processed from the jump host's direct TCP channel and the channel's session object being created in SSHForwardChannel._open_forward(). The channel starts up "paused" as a way to avoid this issue, but it looks like that is currently only being checked for processing data packets, not EOF. So, it's possible for the problematic code in _flush_recv_buf() create be run when the channel is in the 'starting' _recv_paused state. That still seems a bit odd, since the confirmation of the channel open should come in before the EOF message, and that confirmation message should unblock _open_forward(), but perhaps there's still some chance those two coroutines could be scheduled out of order.

Can you confirm the order of the "Received EOF" and "Reading from channel started" messages (if you even see the latter) on the tunnel connection?

For additional confirmation this is the problem, you could add a debug message right before the self._session assignment in SSHForwardChannel._open_forward():

```python
    async def _open_forward(self, session_factory, chantype, *args):
        """Open a forward channel"""

        packet = await super()._open(chantype, *args)

        # Forward channels should have no extra data in the open confirmation
        packet.check_end()

        self.logger.debug('Creating session object')                              <-- new line
        self._session = session_factory()
        self._session.connection_made(self)
        self._session.session_started()
        self._conn.create_task(self._start_reading(), self.logger)

        return self._session
```

If the "Creating" message here happens after "Received EOF", that confirms the problem. It would also be useful to know the order of the "Received EOF" and the "Initial send window" message.

Getting back to your proposed change, the consequence of that would be to not send an EOF message back to the when processing the incoming EOF. That's a safe thing to do for any forwarded connections like what you have here, as they default to supporting "half-open" connections. The same goes for any high-level "stream" API connections. However, this could end up changing the behavior for someone using the low-level callback-based API if they didn't override the SSHSession's default behavior of turning half-closed sessions into fully-closed ones without the application code needing to send EOF explicitly. I'm not sure how much of a problem that would be, but without a session object being created, there's no way to know which behavior should apply.

The good news is that all the places that create a session already call start_reading() after doing so, and that this will end up calling _flush_recv_buf() again. So, I think the right answer here might be to simply not make the transition from 'eof_pending' to 'eof' if we're in the 'starting' state. In other words, I'm thinking of a change such as:

```patch
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -291,7 +291,7 @@ class SSHChannel(SSHPacketHandler):
         while self._recv_buf and not self._recv_paused:
             self._deliver_data(*self._recv_buf.pop(0))
 
-        if not self._recv_buf:
+        if not self._recv_buf and self._recv_paused != 'starting':
             if self._encoding and not exc and \
                     self._recv_state in ('eof_pending', 'close_pending'):
                 try:
```

This should avoid trying to access the session object before it is created, but eventually still end up making the right decision about whether a received EOF should trigger an outbound write_eof() or not. The channel will remain in the 'eof_pending' state for a bit longer, but run this code again as soon as _start_reading() is called, right after creating the session.

This also has the benefit of delaying other logic with regard to trying to close the connection, which might end up triggering other race conditions.

I'm running regression tests on this code right now. Would you be able to test it on your side with the above change in place and your previous change reverted to see if solves your problem?

ashafer01:
Here's a more complete debug log from an unpatched run. Besides the problem host (*.4.15), there may be 1-2 other normal/working connections open using the same tunnel -- this is as minimal as I can get it and have my application still run for testing, but I think those are still making some noise with additional channels; I'm not sure which channel is which from reading the logs:

```
INFO:asyncssh:[conn=0] Opening SSH connection to 172.17.4.15, port 22 via SSH tunnel
INFO:asyncssh:[conn=0] Opening direct TCP connection to 172.17.4.15, port 22
INFO:asyncssh:[conn=0]   Client address: dynamic port
DEBUG:asyncssh:[conn=0, chan=2] Set write buffer limits: low-water=16384, high-water=65536
DEBUG:asyncssh:[conn=0, chan=2]   Initial recv window 2097152, packet size 32768
INFO:asyncssh:[conn=0, chan=1] Received channel close
INFO:asyncssh:[conn=0, chan=1] Channel closed
DEBUG:asyncssh:[conn=0, chan=2]   Initial send window 2097152, packet size 32768
DEBUG:asyncssh:[conn=0, chan=2] Received EOF
DEBUG:asyncssh:[conn=0] Uncaught exception
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 910, in data_received
    while self._inpbuf and self._recv_handler():
  File "/usr/local/lib/python3.8/site-packages/asyncssh/connection.py", line 1143, in _recv_packet
    processed = handler.process_packet(pkttype, seq, packet)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/packet.py", line 215, in process_packet
    self._packet_handlers[pkttype](self, pkttype, pktid, packet)
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 553, in _process_eof
    self._flush_recv_buf()
  File "/usr/local/lib/python3.8/site-packages/asyncssh/channel.py", line 307, in _flush_recv_buf
    not self._session.eof_received() and
AttributeError: 'NoneType' object has no attribute 'eof_received'
INFO:asyncssh:[conn=0] Connection failure: 'NoneType' object has no attribute 'eof_received'
INFO:asyncssh:[conn=0, chan=2] Closing channel due to connection close
INFO:asyncssh:[conn=0, chan=2] Aborting channel
INFO:asyncssh:[conn=0, chan=2] Channel closed: 'NoneType' object has no attribute 'eof_received'
INFO:asyncssh:[conn=3] Connection failure: 'NoneType' object has no attribute 'eof_received'
INFO:asyncssh:[conn=0] Closing connection
INFO:asyncssh:[conn=0] Sending disconnect: Disconnected by application (11)
```

The remainder of the application's output from there is print statements of mine, followed by the usual un-awaited exception asyncio debug output. I don't seem to be getting `Reading from channel started`. This is the only log message I'm seeing for `conn=3` (the expected connection number for the broken host with the current configuration):

```
INFO:asyncssh:[conn=3] Connection failure: 'NoneType' object has no attribute 'eof_received'
```

I also noticed another very strange pattern while trying to minimize the config to reduce log noise: With three connections using the tunnel, this original exception occurs _almost every time_, but not always. Rarely, I get `asyncssh.misc.ConnectionLost: Connection lost`. *However,* with only _two_ connections using the tunnel (the absolute minimum I can do to run the application), the behavior seems to reverse. Every time so far that I've ran with 2 connections using the tunnel, I get `Connection lost`, and 0 of the initial exception in the dozen or so test runs I've tried.

Running with your proposed patch seems to normalize this behavior, and it looks like I'm getting `Connection lost` every time with the 3-connection config as well as with 2 connections. I don't get any timeouts or delays.

I've also verified with your patch in place that other connections can still use the tunnel after this failure occurs (not possible unpatched).

Seems like a better solution as far as I'm concerned; no waiting!

ronf:
Thanks for the more detailed debug log. It looks like the "Received EOF" message on channel 2 is coming in after the "Initial send window" message, as expected, but there's no sign of the "Reading from channel started" message that we'd normally expect to see first. So, it's likely that the EOF is in fact being processed before the code in _open_forward() gets a chance to create the session object, as suspected.

It doesn't surprise me too much that this might be sensitive to the number of parallel tasks you are running, since this is an issue of what order tasks get scheduled in. Since futures are involved, we can't count on run-to-completion when creating the session object, so it really is important to make sure that I/O is fully paused until that session creation occurs...

The regression testing looks good after this change, so I think I'm going to go ahead and check this in (along with a small tweak to a unit test to maintain full code coverage). Thanks for reporting this and proposing a fix, and the effort you've put into helping me track down the root cause. I really appreciate it!

Regarding the Connection Lost messages, it would make sense that you'd see that instead of the AttributeError in cases where you have the fix. Any time an attempted connect() seeing the underlying transport (in this case a channel on the tunnel connection) close before the handshake completes, it will raise this error. The 'Connection Lost' version of the message means you just got a clean close of the underlying transport with no other error (such as when EOF is sent on the channel due to the remote TCP connection closing cleanly).

ronf:
The fix proposed above is now checked into the "develop" branch as commit e9fe32c, and will be rolled into the next release. If you run into any further issues, please let me know. Thanks again for your help!

ashafer01:
Thanks! Also I really appreciate the deep dive here, very enlightening all around.

When would you normally release a patch like this? Since it's breaking the tunnel it's a pretty bad blocker on my end. I'd prefer to avoid depending on a fork in production, but that is definitely an option if you'd prefer not to release right away.

ronf:
Normally, I batch a handful of changes up in "develop" if the issues aren't too severe or the person reporting them is ok with just checking out directly from "develop". However, if you would prefer an official release, I can do that. I actually did a patch release last weekend, only a week after the 2.4.0 feature release, and can probably get another patch out by this weekend if you need it.

ashafer01:
A release before Monday would be absolutely amazing and deeply appreciated

ronf:
Ok - I'll let you know when it's up.

ronf:
This fix is now available in AsyncSSH 2.4.2.

PR #316, opened 2020-09-20
fix #315 - RuntimeError when closing connection

### LINKED ISSUE #315: "RuntimeError: dictionary changed size during iteration" in SSHConnection._cleanup()
If running with Python 3, using `asyncssh.connect(...)` as a context manager, the _cleanup() call will raise `RuntimeError: dictionary changed size during iteration` if there are any local listeners to close.

This is due to a change in Python 3 regarding `Dict.values()` which now returns a [view object](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects) instead of a copy of the dictionary values [as in python 2](https://docs.python.org/2/library/stdtypes.html#dict.values)

Related code : https://github.com/ronf/asyncssh/blob/v2.4.2/asyncssh/connection.py#L584

the fix is to change :

```python
        for chan in list(self._channels.values()):
            chan.process_connection_close(exc)

        for listener in self._local_listeners.values():
            listener.close()
```

with :

```python
        for chan in list(self._channels.values()):
            chan.process_connection_close(exc)

        for listener in list(self._local_listeners.values()):
            listener.close()
```

This bug is present [since v0.2.0](https://github.com/ronf/asyncssh/commit/4be2b0ec90e3003c04b5ce2ab483f41e24d8b43d#diff-9e60bbfe4d79a23a44d680c5674f2b1eR149)

thomasleveil:
"RuntimeError: dictionary changed size during iteration" in SSHConnection._cleanup()

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -581,7 +581,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         for chan in list(self._channels.values()):
             chan.process_connection_close(exc)
 
-        for listener in self._local_listeners.values():
+        for listener in list(self._local_listeners.values()):
             listener.close()
 
         while self._global_request_waiters:


ronf:
This change and some other similar needed changes are now included in commit 89fd910 in the "develop" branch. Thanks very much!

PR #363, opened 2021-04-19
Add client config support for HostKeyAlias

pritambaral:
This adds support for HostKeyAlias client config option, and a
corresponding host_key_alias connection option, to support aliasing host
names in known_hosts and host certificate validation. This is modeled
after [the equivalent OpenSSH feature, as can be seen in ssh_config(5)](https://man.openbsd.org/ssh_config#HostKeyAlias).

----

Since this was a simple enough change in code, and because the functionality is already well-defined based on established usage in OpenSSH and established conventions in AsyncSSH, I chose to avoid creating a separate issue for discussing this. I think any needed discussion can happen here just as well; any required code changes can be easily carried out right here.

My personal use case for this config option is to be able to connect to servers whose IP address(es) cannot be resolved from their hostnames, without having to drop the security of host name verification or resorting to ugly hacks with per-host UserKnownHostsFile. To illustrate:

```sh
$ ping ssh-test-host
ping: ssh-test-host: Temporary failure in name resolution

$ cat <<EOF >~/.ssh/config
Host ssh-test-host
  Hostname 10.10.10.10
EOF

$ ssh ssh-test-host
The authenticity of host '10.10.10.10 (10.10.10.10)' can't be established.
ED25519 key fingerprint is ...
Are you sure you want to continue connecting (yes/no/[fingerprint])? 
```

See how the hostname (and IP address) ssh(1) uses to authenticate the server are both just the IP address? This is because ssh(1) uses the value of `Hostname` in the config as the 'host' for the purposes of authentication. This leaves us with IP-based host key checking as the only way to check host keys, which is undesirable for regular host key authentication but also renders hostname-based certificate authentication useless.

But, by setting `HostKeyAlias` to the host name that we *want* to be used for the purposes of host authentication:

```sh

$ cat <<EOF >~/.ssh/config
Host ssh-test-host
  Hostname 10.10.10.10
  HostKeyAlias ssh-test-host
EOF

$ ssh ssh-test-host
The authenticity of host 'ssh-test-host (10.10.10.10)' can't be established.
ED25519 key fingerprint is ...
Are you sure you want to continue connecting (yes/no/[fingerprint])? 
```

ssh(1) now uses a proper host name, instead of IP, for host auth.

### DIFF
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -481,6 +481,7 @@ class SSHClientConfig(SSHConfig):
         ('HostbasedAuthentication',         SSHConfig._set_bool),
         ('HostKeyAlgorithms',               SSHConfig._set_string),
         ('Hostname',                        _set_hostname),
+        ('HostKeyAlias',                    SSHConfig._set_string),
         ('IdentitiesOnly',                  SSHConfig._set_bool),
         ('IdentityAgent',                   SSHConfig._set_string),
         ('IdentityFile',                    SSHConfig._append_string),
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -2553,6 +2553,7 @@ class SSHClientConnection(SSHConnection):
         self._port = options.port
 
         self._known_hosts = options.known_hosts
+        self._host_key_alias = options.host_key_alias
 
         self._server_host_key_algs = options.server_host_key_algs
         self._server_host_key = None
@@ -2634,7 +2635,8 @@ class SSHClientConnection(SSHConnection):
 
             port = self._port if self._port != DEFAULT_PORT else None
 
-            self._match_known_hosts(self._known_hosts, self._host,
+            self._match_known_hosts(self._known_hosts,
+                                    self._host_key_alias or self._host,
                                     self._peer_addr, port)
 
         default_host_key_algs = []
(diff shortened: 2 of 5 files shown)


ronf:
Thanks very much for your contribution! This looks like a useful feature, and your implementation looks nice and clean. Can you confirm that you're ok providing these changes under the current AsyncSSH dual-license (EPL-2.0 or GPL 2.0 or later)?

The only thing missing here is unit tests, both for the new host_key_alias argument and the config option, and also that the known_hosts matching properly uses the new host_key_alias in the known hosts matching. It's pretty clear in this case from looking at the change that it should, but it's good to have tests in place for that to catch any regressions caused by future commits.

pritambaral:
> This looks like a useful feature, and your implementation looks nice and clean.

Possible only because the code-base was already nice and clean.

> Can you confirm that you're ok providing these changes under the current AsyncSSH dual-license (EPL-2.0 or GPL 2.0 or later)?

Yes, I agree to providing these licenses under the terms of the Eclipse Public License v2.0 which is available at http://www.eclipse.org/legal/epl-2.0/ and also under the GNU General Public License, Version 2.0, or any later versions of that license, when the conditions for such availability set forth in the Eclipse Public License v2.0 are satisfied.

> The only thing missing here is unit tests ...

Tests have been added. They're not as clean as I'd have liked them to be, but I couldn't make them cleaner without modifying existing code or breaking existing conventions (further).

ronf:
Thanks very much! I really appreciate your contribution...

Regarding the unit tests, I went ahead and updated the parent class connect() function so that it was usable for this testing, and made some other very minor tweaks.

The result is now checked in as commit fabacc3 in the "develop" branch, and will be included in the next release!

PR #366, opened 2021-05-08
Fixed arguments for pty_requested session callback

velavokr:
`self._term_*` fields were passed to `pty_requested` callback before assignment. I suppose the intention was to pass the corresponding local variables.

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -1428,8 +1428,8 @@ class SSHServerChannel(SSHChannel):
             else:
                 raise ProtocolError('Invalid pty modes string')
 
-        result = self._session.pty_requested(self._term_type, self._term_size,
-                                             self._term_modes)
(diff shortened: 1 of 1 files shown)


ronf:
Thank you for the report! It looks like using the member variables was correct originally, but that was broken by commit b6f5e7d back in 2016 and it was never noticed. That change unpacked values into local variables and only copied them into the member variables when pty_requested() returned that a PTY was allowed. However, as you discovered, that callback was not updated to use the local variables. It didn't crash because the member variables were initialized, but not to the values that should have been passed to the callback.

I'm going to fix this in the develop branch and add a unit test associated with it, and it'll be merged to master as part of the next release.

ronf:
This is now fixed in commit 7f323c3 in the "develop" branch. Thanks again!

ronf:
This fix is now available in AsyncSSH 2.7.0.

PR #367, opened 2021-05-12
Added a get_terminal_modes method to SSHServerChannel and SSHServerProcess

velavokr:
This is to get all the modes without iterating over all the possible keys or going for nonpublic fields.

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -1687,6 +1687,27 @@ class SSHServerChannel(SSHChannel):
 
         return self._term_modes.get(mode)
 
+    def get_terminal_modes(self):
+        """Return all TTY modes for this session
+
+           This method returns all the values of POSIX terminal modes
+           set by the client when the session was opened. If the client
+           didn't request a pseudo-terminal or didn't set the requested
+           TTY mode opcode, this method will return an empty dictionary.
+           Calls to this method should only be made after
+           :meth:`session_started <SSHServerSession.session_started>`
+           has been called on the :class:`SSHServerSession`. When using
+           the stream-based API, calls to this can be made at any time
+           after the handler function has started up.
+
+           :returns: A dictionary containing all the requested POSIX
+                     terminal modes or an empty dictionary if no modes
+                     were requested
+
+        """
+
+        return self._term_modes
+
     def get_x11_display(self):
         """Return the display to use for X11 forwarding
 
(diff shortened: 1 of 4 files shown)


ronf:
Can you say a bit more about what you are trying to do here? I'm hesitant to directly return the `_term_modes` dictionary because a caller could then modify that dictionary. I could return a copy of the dictionary to avoid that, but depending on your use case there might be other options that make sense, like returning an iterator or dictionary view.

ronf:
Thanks for providing more detail.

Looking into this a bit further, it seems like types.MappingProxyType() would be a good choice for this. It basically returns a read-only version of a dictionary without paying the cost to copy it. It's used internally for things like the dictionary associated with Python classes. Here's an article I found which discusses it further:

https://medium.com/python-features/all-we-need-to-know-about-various-dictionaries-in-python-3564d6231f20

With this, all the usual operations on a dictionary should work except for anything that tries to modify it, and the original dictionary won't be at risk of being changed.

I'll look into adding this to the "develop" branch, along with some unit tests.

velavokr:
Great! I'll update my PR then.

velavokr:
Updated the relevant tests.

ronf:
Thanks for updating the PR. I had already started on making some changes locally after your first submission including some updates to the unit tests, but I have gone back and incorporated some of what you submitted here.

After looking at the SSHServerProcess code, I did take one thing in a bit of a different direction. Unlike SSHChannel,  most of the values associated with SSHProcess classes are available as properties rather than get_*() functions. I decided to use this as an opportunity to update the terminal values to also follow this pattern. So, there are now new properties named term_type, term_size, and term_modes in SSHServerProcess, and you can use term_modes to do what you want here, or you can call term_modes.get() to get the equivalent of the previous get_terminal_mode() functionality.

For backward compatibility, the old get_terminal_*() functions are still present of course, but unlike SSHServerChannel I didn't add a get_terminal_modes() function to SSHServerProcess. The preferred mechanism for accessing both the complete list of modes and looking up individual modes is now through the properties. At some point in the future I may drop support for the get_terminal_*() functions and support only the properties, but probably not until the next major release.

These changes are now available in commit e340d14 in the "develop" branch. Let me know if you see any issues.

velavokr:
Thanks! Looks great!

velavokr:
Can I drop the PR then?

ronf:
Yeah - feel free to go ahead and close this if my version of the changes look ok.

I'll also be including this in the main branch as part of the next release.

ronf:
This feature is now available in AsyncSSH 2.7.0.

PR #381, opened 2021-07-06
Fixed some Security Issues - Observable Discrepancy leading to an information leak in the algorithm negotiation and trivial success authentication

manfred-kaiser:
AsyncSSH has an Observable Discrepancy leading to an information leak in the algorithm negotiation. This allows man-in-the-middle attackers to target initial connection attempts (where no host key for the server has been cached by the client). 

This vulnerability allows a man in the middle attack to determine, if a client already has prior knowledge of the remote hosts fingerprint.

Using this information leak it is possible to ignore clients, which will show an error message during an man in the middle attack, while new clients can be intercepted without alerting them of the man in the middle attack.

This is the same vulnerability like in [OpenSSH](https://docs.ssh-mitm.at/CVE-2020-14145.html) and [PuTTY](https://docs.ssh-mitm.at/CVE-2020-14002.html).

This pull request introduces a new flag, which enables a strict security mode. With this mode enabled, some common mitm attacks are harder to achieve.

Client example:

```python
import asyncio, asyncssh, sys

asyncssh.connection._STRICT_MODE = True

async def run_client():
    async with asyncssh.connect('localhost' , port=10022) as conn:
        result = await conn.run('ls', check=True)
        print(result.stdout, end='')

try:
```

Example known host fingerprint:

```
['rsa-sha2-256', 'rsa-sha2-512', 'ssh-rsa-sha224@ssh.com', 'ssh-rsa-sha256@ssh.com', 'ssh-rsa-sha384@ssh.com', 'ssh-rsa-sha512@ssh.com', 'ssh-rsa']
```

Without known fingerprint or with `asyncssh.connection._STRICT_MODE = True`

```
['sk-ssh-ed25519-cert-v01@openssh.com', 'sk-ecdsa-sha2-nistp256-cert-v01@openssh.com', 'ssh-ed25519-cert-v01@openssh.com', 'ssh-ed448-cert-v01@openssh.com', 'ecdsa-sha2-nistp521-cert-v01@openssh.com', 'ecdsa-sha2-nistp384-cert-v01@openssh.com', 'ecdsa-sha2-nistp256-cert-v01@openssh.com', 'ecdsa-sha2-1.3.132.0.10-cert-v01@openssh.com', 'ssh-rsa-cert-v01@openssh.com', 'sk-ssh-ed25519@openssh.com', 'sk-ecdsa-sha2-nistp256@openssh.com', 'ssh-ed25519', 'ssh-ed448', 'ecdsa-sha2-nistp521', 'ecdsa-sha2-nistp384', 'ecdsa-sha2-nistp256', 'ecdsa-sha2-1.3.132.0.10', 'rsa-sha2-256', 'rsa-sha2-512', 'ssh-rsa-sha224@ssh.com', 'ssh-rsa-sha256@ssh.com', 'ssh-rsa-sha384@ssh.com', 'ssh-rsa-sha512@ssh.com', 'ssh-rsa']
```

If strict mode is enabled, the default algorithm list will be sent to the server and a mitm attacker does not know, if this is the first connection attempt or if strict mode is enabled.

If a mitm server wants to intercept the connection and the server fingerprint is known, asyncssh will close the session because of different fingerprints.

If strict mode is not enabled and the server detects, that the client allready knows the fingerprint, the mitm server can pass the connection to the destination server. With this method, a mitm server avoids raising errors due to invalid host fingerprints.

### DIFF
--- a/asyncssh/auth.py
+++ b/asyncssh/auth.py
@@ -132,6 +132,7 @@ class _ClientGSSKexAuth(_ClientAuth):
             self.logger.debug1('Trying GSS key exchange auth')
 
             await self.send_request(key=self._conn.get_gss_context())
+            self._conn.trivial_auth = False
         else:
             self._conn.try_next_auth()
 
@@ -184,7 +185,7 @@ class _ClientGSSMICAuth(_ClientAuth):
             token = self._gss.step()
 
             self.send_packet(MSG_USERAUTH_GSSAPI_TOKEN, String(token))
-
+            self._conn.trivial_auth = False
             if self._gss.complete:
                 self._finish()
         except GSSError as exc:
@@ -306,6 +307,7 @@ class _ClientPublicKeyAuth(_ClientAuth):
 
         self.logger.debug1('Signing request with %s key',
                            self._keypair.algorithm)
+        self._conn.trivial_auth = False
 
         await self.send_request(Boolean(True),
                                 String(self._keypair.algorithm),
@@ -381,6 +383,7 @@ class _ClientKbdIntAuth(_ClientAuth):
         num_prompts = packet.get_uint32()
         prompts = []
         for _ in range(num_prompts):
+            self._conn.trivial_auth = False
             prompt = packet.get_string()
             echo = packet.get_boolean()
 
@@ -422,7 +425,7 @@ class _ClientPasswordAuth(_ClientAuth):
             return
 
         self.logger.debug1('Trying password auth')
-
+        self._conn.trivial_auth = False
         await self.send_request(Boolean(False), String(password))
 
     async def _change_password(self, prompt, lang):
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -169,6 +169,8 @@ _DEFAULT_MAX_PKTSIZE = 32768        # 32 kiB
 _DEFAULT_LINE_HISTORY = 1000        # 1000 lines
 _DEFAULT_MAX_LINE_LENGTH = 1024     # 1024 characters
 
+_STRICT_MODE = False
+
 
 async def _open_proxy(loop, command, conn_factory):
     """Open a tunnel running a proxy command"""
@@ -2792,6 +2794,10 @@ class SSHClientConnection(SSHConnection):
             if self._x509_trusted_certs or self._x509_trusted_cert_paths:
                 default_host_key_algs = \
                     get_default_x509_certificate_algs() + default_host_key_algs
+
+        if _STRICT_MODE:
+            default_host_key_algs = \
+                get_default_certificate_algs() + get_default_public_key_algs()
 
         self._server_host_key_algs = \
             _select_host_key_algs(
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -667,6 +667,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         self._auth_complete = False
         self._auth_methods = [b'none']
         self._username = None
+        self.trivial_auth = True
 
         self._channels = {}
         self._next_recv_chan = 0
@@ -2011,7 +2012,9 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
 
     def _process_userauth_success(self, _pkttype, pktid, packet):
         """Process a user authentication success response"""
-
+        if _STRICT_MODE and self.trivial_auth:
+            self._force_close(PermissionDenied('trivial auth'))
+            return
         packet.check_end()
 
         if self.is_client() and self._auth:


manfred-kaiser:
Fixed a second vulnerability, which allows an attacker to spoof the login process and force a client to login to a server.

More information can be found in the commit message of PuTTY: https://git.tartarus.org/?p=simon/putty.git;a=commit;h=1dc5659aa62848f0aeb5de7bd3839fecc7debefa

ronf:
Thank you for the contribution. I'm not sure I fully understand the change you're trying to make, but I don't think the code submitted here does what you intend. Right now, the first change would overwrite the advertised server host key algorithms to include all possible algorithms independent of any algorithms associated with the server host keys that match in known_hosts, but it does not override other ways the client can set an explicit set of server host key algs using the options or config mechanisms. Is that what you intended? In fact, using those config mechanism already provides a way to advertise any list of host key algorithms you choose, independent of what  algorithms are matched in the known_hosts lookup.

Note also that setting all possible known host key algorithms still makes it possible to fingerprint the client, since AsyncSSH supports a much larger number of mechanisms by default than most other SSH clients. It also reports itself in the initial banner string like many other SSH clients. While it is a host debate topic whether that's a good or bad thing, I tend to be on the side of obfuscating the client version doing more harm than good.

The second change makes more sense to me, but I wouldn't want this done by globally setting a private variable deep in the AsyncSSH implementation. This should be made a proper client connection option, settable on a per connection basis in the connect() call. I can look into doing this when I pull in the change.

manfred-kaiser:
> Thank you for the contribution. I'm not sure I fully understand the change you're trying to make, but I don't think the code submitted here does what you intend. Right now, the first change would overwrite the advertised server host key algorithms to include all possible algorithms independent of any algorithms associated with the server host keys that match in known_hosts, but it does not override other ways the client can set an explicit set of server host key algs using the options or config mechanisms. Is that what you intended? In fact, using those config mechanism already provides a way to advertise any list of host key algorithms you choose, independent of what algorithms are matched in the known_hosts lookup.

The intention is to set the ordering of hostkey algorithms to the default list, but only when a configuration option is set.
If a user/developer wants to use another list, than this is not part of this patch. In such cases, I assume, that the developer has some reasons for using another ordering.

> Note also that setting all possible known host key algorithms still makes it possible to fingerprint the client, since AsyncSSH supports a much larger number of mechanisms by default than most other SSH clients. It also reports itself in the initial banner string like many other SSH clients. While it is a host debate topic whether that's a good or bad thing, I tend to be on the side of obfuscating the client version doing more harm than good.

I think that fingerprinting the client is not a problem, because the client name and version is sent to the server. If you want to detect specific clients, this is far easier and obfuscating the client version string can lead to other problems. Using the default host key algorithm has nothing to do with obfuscation. This fix makes it harder for a man in the middle attacker to find vulnerable targets.

PuTTY and OpenSSH has implemented two different approaches to mitigate this vulnerability.

In PuTTY you can enable a configuration option to send the default ordering for host key algorithms. This is the same mitigation approach as implemented in this patch. 

OpenSSH's implementation is more sophisticated. Explanation from Damien Miller:

```
That being said, commit b3855ff (shipped in openssh-8.4) adjusted the ordering to always 
use the default if the client has learned a hostkey matching the best-preference algorithm. 
openssh-8.5 enabled UpdateHostkeys by default (with some restrictions) so most users will 
automatically learn a best-preference hostkey if one is available at the server. Between these, 
most users should end up using the default algorithm list.
```


> The second change makes more sense to me, but I wouldn't want this done by globally setting a private variable deep in the AsyncSSH implementation. This should be made a proper client connection option, settable on a per connection basis in the connect() call. I can look into doing this when I pull in the change.

I used this private variable, because both fixes check if this variable is set. As you mentioned, this approach is not best practice, but I don't wanted to change too much.

If you are interested, I can send you more information about the trivial success authentication vector. At the moment I want to fix this vulnerability in as many clients as possible. So there is no full disclosure available about this attack vector.

Can I send you a mail to the mail address in you git log?

manfred-kaiser:
I have sent you a mail with a complete description and a PoC of the trivial success authentication attack vector.

ronf:
Thanks - I'll take a look and respond via e-mail.

ronf:
Here's a patch to address the first issue around known_hosts:

```patch
@@ -419,7 +419,7 @@ def _select_algs(alg_type, algs, possible_algs, default_algs,
     else:
         strict_match = True

-    if algs == ():
+    if algs in ((), 'default'):
         return default_algs
     elif algs:
         if isinstance(algs, str):
@@ -451,10 +451,6 @@ def _select_host_key_algs(algs, config_algs, default_algs=()):
     possible_algs = (get_x509_certificate_algs() + get_certificate_algs() +
                      get_public_key_algs())

-    if not default_algs:
-        default_algs = (get_default_certificate_algs() +
-                        get_default_public_key_algs())
-
     return _select_algs('host key', algs, possible_algs,
                         default_algs, config_algs)

@@ -2777,7 +2773,8 @@ class SSHClientConnection(SSHConnection):

         default_host_key_algs = []

-        if self._trusted_host_key_algs:
+        if self._trusted_host_key_algs and \
+                self._server_host_key_algs != 'default':
             default_host_key_algs = self._trusted_host_key_algs

         if self._trusted_ca_keys:
```

To activate it, pass in the argument server_host_key_algs='default' in your asyncssh.connect() call. It should continue to do the known_hosts validation but send the default list of host key algorithms in the handshake that you'd normally only see when passing in known_hosts=None.

If this looks good to you, let me know and I can check this in and move on to implementing a new 'strict_auth' option based on your other proposed changes.

manfred-kaiser:
Your solution with using 'default' is very good :-D

Using "default" as argument makes it easy to use.

If you want to check your changes, you can use my ssh-mitm tool. I have added a check for asyncssh to the develop branch.

[https://github.com/ssh-mitm/ssh-mitm](https://github.com/ssh-mitm/ssh-mitm)

You can start ssh-mitm without any arguments. This will start ssh-mitm on port 10022

```
ssh-mitm
```

If you don't have a known fingerprint or the default order is used, you will see following log message:

```
asyncssh: Client connecting for the FIRST time!
```


If you have a known fingerprint and default order is not used, following message is shown:

```
asyncssh: Client has a locally cached remote fingerprint!
```

ronf:
I did a bit more testing of this change tonight. While everything is working as expected, I ran into something I hadn't considered before that might be a problem. If you advertise all default server host key algorithms, you run the risk of the server picking a host key which the client doesn't trust. This isn't a problem when the default algorithms are sent when known_hosts is set to None, as all host keys sent by the server are trusted in that case (as long as the algorithm is something the client supports). However, if you have a known_hosts file that only trusts an RSA key but you send the full set of default algorithms and get back an ECDSA key not in known_hosts, that leads to a connection failure.

Unlike public-key client auth, the server is only allowed to send a single host key, so this is a problem. I'm starting to think allowing additional algorithms to be sent in server_host_key_algs while leaving known_hosts enforcement enabled is not a good idea.

manfred-kaiser:
> If you advertise all default server host key algorithms, you run the risk of the server picking a host key which the client doesn't trust. 

**This is an ecpexted behavior.** When using only default algorithms, there is the risk, that the host key is untrusted.
I have mentioned this problem in [https://docs.ssh-mitm.at/CVE-2020-14145.html](https://docs.ssh-mitm.at/CVE-2020-14145.html)

You should only use default host keys for a specific host, if the risk of a mitm attack by leaking this information is higher than the problem with untrusted host keys.

As Damien Miller mentioned, for most users, presenting an untrused host key will confuse them, which can result in a higher risk of mitm attacks, if the users does not verify it. This is the reason, why OpenSSH has chosen a more user orientated approach to lower the risk of this information leak. The difference is, that OpenSSH is a product for endusers and most of them does not care about implementation details.

> I'm starting to think allowing additional algorithms to be sent in server_host_key_algs while leaving known_hosts enforcement enabled is not a good idea.

It depends on the use case. For most scenarios you are right and I don't recommend using default algorithms for every connection. You should use this option only for specific connections.

I think it's up to the application developer to decide to use the default algorithms or the preferred algorithm order. With the "default" keyword there is an easy option to use this feature and a application developer must not know the default algorithms, which can also change from version to version, resulting in the same information leak as before.

manfred-kaiser:
You have asked for the yonfiguration option to disable tirival success authentications.

Dropbear will implement following configuration option: ``-o DisableTrivialAuth=yes``

If OpenSSH decides to use my path, this will be the same config option. But as I mentioned in the mail, I think that OpenSSH will prefere a different approach to mitigate the risk of the trivial success authentication attack vector.

OpenSSH has also the problem, that many Linux Distributions uses an old version of OpenSSH. This is one reason, why a different approach is needed. Damien Miller is working on a new ssh agent, which will also mitigate this attack vector: https://github.com/djmdjm/openssh-wip/pull/3 and https://github.com/djmdjm/openssh-wip/pull/4

ronf:
Thanks for the additional information -- while there are some risks about getting a non-matching server host key, I've gone ahead with the implementation and noted that issue in the documentation for the new option.

The server_host_key_algs change is now in the "develop" branch as commit 90d7590 and the disable_trivial_auth change is commit ed6e0f2. I took a slightly different approach on the trivial auth change to pass the "trivial auth" information via arguments to calls made into the connection object during auth rather than directly setting state on the connection object. The end result should be equivalent, though.

The specific forms of auth which are treated as "non-trivial" are:

- GSS Kex auth when it signs a userauth request with the key from the GSS key exchange
- GSS MIC auth when it signs the MIC with a key associated with the GSS token
- Host-based auth when it signs a userauth request with the client host key
- Public key auth when it signs a userauth request with the client key
- Password auth when it provides a password or performs a password change
- Keyboard-interactive auth when it sends a response to a non-empty list of prompts

The forms of auth treated as "trivial" include the server returning success on the very first client request to retrieve a list of auth methods and negotiating keyboard-interactive auth but sending only empty lists of prompts, so the client doesn't actually get to provide any response.

These two changes also include enhance unit tests, checking the specific host key algorithms which are advertised in various cases and checking that disable_trivial_auth properly allows or denies different forms of authentication.

Let me know if you see any problems with either of these changes. I'll wait a bit before moving this code into a new release. Thanks again for contacting me about this!

manfred-kaiser:
Your implementation for mitigating triviall success auth looks good, but you have added "Host-based auth" as a non trivial authentication. As there are some security concerns regarding this authentication type, we not recommend treating host based auth as non trivial.

ronf:
I debated about this, but I wasn't sure that the weaknesses people normally attribute to host-based auth impacted this specific issue. Is this because the client wouldn't be challenged even when it is asked to do the signing with client host key? Since there isn't any form of agent forwarding applicable here to pass that authentication on to the actual server, I couldn't see a reason to make that subject to this new flag. If you'd prefer to discuss the details in e-mail rather than here, we can do that.

Note that there already is a separate config option to enable/disable host-based auth, and it is only enabled by default if the client machine makes the host keys available (usually by installing a setuid "keysign" executable).

On a separate note, I was also wondering if I should reference https://nvd.nist.gov/vuln/detail/CVE-2020-14145 specifically in the documentation for this feature. What would you recommend?

manfred-kaiser:
I have sent you a mail

PR #383, opened 2021-07-08
Use funtools.partial instead of global in remote forwarding example

termim:
This is kind of nitpicking but usage funtools.partial is more in spirit with asyncio documentation.


ronf:
Thanks very much - I didn't like using global there, but didn't want to complicate the example too much. This is a nice solution that still keeps it pretty simple!

PR #384, opened 2021-07-15
SSHConfig._include: Parse homedir-relative paths

zmc:
e.g. '~/.ssh/config.d/foo'. These were getting silently ignored before.

Signed-off-by: Zack Cerza <zack@redhat.com>

### DIFF
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -111,7 +111,7 @@ class SSHConfig:
         # pylint: disable=unused-argument
 
         for pattern in args:
-            path = Path(pattern)
+            path = Path(pattern).expanduser()
 
             if path.anchor:
                 pattern = str(Path(*path.parts[1:]))


zmc:
Hi, first-time contributor here. Since this is a one-liner I though an issue would be unnecessary. EPLv2 is fine!

ronf:
Thanks very much for catching and reporting this, Zach... I've brought your fix into the "develop" branch as commit 1559e8f and will be including the change in the next release.

zmc:
Could you explain why you choose not to merge PRs and instead take contributors' work into your own commits, please? I can't recall seeing behavior like this in my years working with open source projects.

ronf:
I generally prefer to check in changes in the "develop" branch, and only move them over to "master" when I'm ready to generate a new release, especially once there are other pending changes already in "develop" that have not made it into "master" yet. A simple bug fix like this could have gone into master directly and I could have re-based "develop" against that, but since I already had "develop" checked out, it was simpler to just make the one-line change there. Sorry if that was unexpected. Whenever I pull in someone else's change, I try to make sure to give them credit in the check-in comment, as I did here.

If you look back at the history, you'll see I have taken in other PRs directly in some cases, but it depends on how complete the change is, and whether I need to make edits to it before bringing it in. A recent example was 43620cb, and a more involved example was fabacc3.

sebastian-philipp:
Hi, did you know that you can also use `git cherry-pick` for taking commits? This way you can keep the original author of the commit in the git commit history.

PR #401, opened 2021-08-30
tests: print failed commands

ktdreyer:
Show the exact command that failed, so that developers can reproduce and debug failing commands more easily.

For example, this makes it easier to understand when tests run `openssl` commands with unsupported options.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -192,6 +192,7 @@ def run(cmd):
         return subprocess.check_output(cmd, shell=True,
                                        stderr=subprocess.STDOUT)
     except subprocess.CalledProcessError as exc: # pragma: no cover
+        print('command failed: %s' % cmd)
         print(exc.output.decode())
         raise
 


ronf:
The print() statement there was actually a bit of debugging that got checked in accidentally some time ago. I noticed it later, but never got around to doing this properly through the 'logging' system rather than as a `print()` statement. I like your suggestion to add the command which failed, but this seems like a good opportunity to also finally get around to replacing that print() with a proper logging call.

ronf:
After putting out release 2.7.1, I realized I had forgotten to include this change. As noted above, I decided to replace the print() with a logger call:

```patch
diff --git a/tests/util.py b/tests/util.py
index 5ef8cb1..b947fe5 100644
--- a/tests/util.py
+++ b/tests/util.py
@@ -192,7 +192,8 @@ def run(cmd):
         return subprocess.check_output(cmd, shell=True,
                                        stderr=subprocess.STDOUT)
     except subprocess.CalledProcessError as exc: # pragma: no cover
-        print(exc.output.decode())
+        logger.error('Error running command: %s' % cmd)
+        logger.error(exc.output.decode())
         raise

```

Sorry for missing it in the release I just put out, but this change is now available in commit be8af38 in the "develop" branch and will be included in the next release.

Issue #415, opened 2021-10-27
login_timeout not applying to asyncssh.connect()


jgauth:
Python version: 3.8.9
asyncssh version: 2.7.2

I'm calling asyncssh.connect() with login_timeout=30, but it is still only timing out after the default value of 2 minutes.

Call looks like:
```
async with asyncssh.connect(host, username='user', password='pass', known_hosts=None, login_timeout=30) as conn:
```

Debug logs:
```
DEBUG:asyncio:Using selector: EpollSelector
INFO:asyncssh:Opening SSH connection to redacted_hostname, port 22
DEBUG:asyncio:Get address info redacted_hostname:22, type=<SocketKind.SOCK_STREAM: 1>
DEBUG:asyncio:Getting address info redacted_hostname:22, type=<SocketKind.SOCK_STREAM: 1> took 2.615ms: [(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('redacted_ip', 22))]

< hangs here for 2 minutes >

DEBUG:asyncio:Close <_UnixSelectorEventLoop running=False closed=False debug=True>
```
Let me know if there's anything else I could provide.
Thanks!

ronf:
Hi John,

Thanks for the suggestion on the doc update. I think that's a good idea. Given the multiple questions I've had about this, I decided to take this a step further, though, and add a new connect_timeout connection option which is basically shorthand for calling asyncio.wait_for(). So, you can now do something like:

```python
async with asyncssh.connect(host, connect_timeout=30) as conn:
```
instead of
```python
async with asyncio.wait_for(asyncssh.connect(host), timeout=30)) as conn:
```
I've also now converted the "ConnectTimeout" client config file option to use this new capability, making it better match the behavior of OpenSSH. The config file support also means you can set different connect timeouts based on host or other matching criteria in the config file.

This new option applies any time an outbound connection is being made, even if that connection is used to set up a "reverse" SSH server connection, or even to set up a tunnel which is then used to open an SSH listener. It can also be used when connecting to an SSH server to retrieve host keys from it.

This change is available as commit b2cadab in the "develop" branch if you'd like to give it a try, and I'll be including this in the next release.

jgauth:
Looks great, thanks a lot!

ronf:
My pleasure! This new functionality is now released in AsyncSSH 2.8.0.

Issue #411, opened 2021-10-07
asyncssh.connection.run() does not seem to respect timeout parameter


bkienker:
I'm trying to allow the caller to specify a timeout value for SSH connections with asyncssh. Here's the function that creates the connection:

```
async def _run_client(host, command, timeout, **kwargs):
    async with asyncssh.connect(host, **kwargs) as conn:
        return await conn.run(command, timeout=timeout), host
```

The coroutine is gathered into a list of tasks in another function:

```
async def _run_multiple_clients(hosts, command, timeout, **kwargs):
    tasks = (_run_client(host, command, timeout, **kwargs) for host in hosts)
    results = await asyncio.gather(*tasks, return_exceptions=True)

    return results
```

And the tasks are run with a third function:

```
def run_ssh(hosts, command, timeout=60, **kwargs):
    return asyncio.get_event_loop().run_until_complete(
        _run_multiple_clients(hosts, command, timeout=timeout, **kwargs)
    )
```

In testing,  I've found that when I simulate a connection timeout by attempting to connect to a host that is behind a firewall, my function always takes a little over two minutes before raising the TimeoutError.

Should this timeout parameter affect the connection timeout wait time, or is it just for command execution time after the connection is established?

I'm sure I must be doing something wrong, thank you in advance. Please let me know if any additional information would be helpful.

ronf:
The timeout in the run() method definitely only counts the time to run the command, as it doesn't even begin to execute until after the asyncssh.connect() call returns with a fully established (and authenticated) connection.

Even if you were to try and add a login_timeout to the connect() call, that will not begin to apply until after the SSHClientConnection object is created, which only happens after the TCP connection is established. So, it will allow you to time out if the SSH handshake takes too long but not if the TCP connect is taking a long time, like in your example here with a firewall.

If you want to enforce a timeout which includes the time to make the TCP connection, you'll need to use something like asyncio.wait_for() wrapped around whatever block of code you want the timeout to apply to. Take a look at #409 for more information on this.

bkienker:
OK, that is extremely helpful! Thank you for your quick reply. I will take a look at asyncio.wait_for().

ronf:
After multiple requests/questions about this, I've decided to add a new "connect_timeout" parameter that is basically shorthand for wrapping your connect call in asyncio.wait_for(). See #415 for more details about the new option, which is now available as commit b2cadab in the "develop" branch.

bkienker:
That's excellent, thanks for your work!

PR #416, opened 2021-10-28
Clarify login_timeout documentation

jgauth:
There's been a few issues (#409 #411 #415 ) opened regarding login_timeout only applying after TCP connection. This might help clarify how it works.


ronf:
A variant of this is now included in the "develop" branch as part of commit b2cadab. Thanks for the suggestion!

Issue #376, opened 2021-06-08
[Question] Generalized alternative of ProxyCommand


isidentical:
There is a [`ProxyCommand`](http://docs.paramiko.org/en/stable/api/proxy.html#module-paramiko.proxy) class is paramiko which acts like the `ProxyCommand` directive on the ssh config. 

> This class implements a the socket-like interface needed by the Transport and Packetizer classes. Using this class instead of a regular socket makes it possible to talk with a Popen’d command that will proxy traffic between the client and a server hosted in another machine.

When creating a new connection, we pass it as the socket that is going to use. E.g;
https://gist.github.com/ovidiucs/e0aebf203f9280ead466#file-paramiko-proxy-py-L38

The class basically acts as a socket-like interface and on `send()` calls send the data to the stdin of the started subprocess (which is the proxy command) and on `read()` calls it reads the data from the stdout. See the implementation here;
https://github.com/paramiko/paramiko/blob/d2d3d77395d1079a2e335080b7f9f79d3f70ad2a/paramiko/proxy.py#L71-L119

I thought about implementing a similar class (perhaps using async subprocesses, to make the `send`/`recv` non-blocking) though I don't know if there is anything I should be concerned about (does asyncssh use anything besides a regular `send`/`recv` on the sockets for example?). I've seen your answer on here https://github.com/ronf/asyncssh/issues/104#issuecomment-324818754 which shows how to pass an external socket to proxy the connection over, so I intend to use that example as the base. If you have any suggestions / or stuff that I should keep in mind while implementing, I'd appreciate it.

ronf:
Here's an example of how to connect a SOCKS proxy into AsyncSSH using a pattern similar to #104, but with the "aiosocks" library providing the socket I/O in a form that's compatible with asyncio:

```python
import aiosocks, asyncio, asyncssh, sys

class SOCKSClientConnection:
    def __init__(self, proxy, proxy_auth=None):
        self._proxy = proxy
        self._proxy_auth = proxy_auth

    async def create_connection(self, session_factory, host, port):
        return (await aiosocks.create_connection(session_factory,
                                                 proxy=self._proxy,
                                                 proxy_auth=self._proxy_auth,
                                                 dst=(host, port)))

async def run_client():
    tunnel = SOCKSClientConnection(aiosocks.Socks5Addr('localhost', 1080),
                                   aiosocks.Socks5Auth('socksuser', 'sockspw'))

[...]

PR #429, opened 2021-12-20
Adding a tunnel class to connect to a remote host via HTTP Connect

NalinG:
Usually, you cannot initiate an SSH connection via a non-standard transport( here: HTTP Connect).  This change adds a Protocol class which allows you to ssl to a proxy host and then send an HTTP Connect request to the target host. After that, you can setup an SSH connection over the tunnel. 

I haven't added any tests yet as I wanted to check in with the maintainer about the method and plausibility of adding this to the asyncssh lib.

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -27,6 +27,7 @@ import io
 import os
 import shlex
 import socket
+import ssl
 import sys
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


ronf:
Thank you for the submission, but I'm not sure this code really belongs in AsyncSSH itself. There are MANY different ways to tunnel traffic -- SOCKS, Proxy Protocol, and HTTP CONNECT all come to mind, and that's far from a complete list. Rather than growing the AsyncSSH code to include all of these and more, I tried to provide a general mechanism that would allow a caller to implement tunneling code (much like you did here) in their own application, with AsyncSSH then using that object in place of a standard TCP socket object. I also recently provided a way to allow the tunneling code to run in a separate process, for cases where someone wanted to tunnel using code not written in Python.

You can find an example of creating a SOCKS tunnel and some other related discussion of the more generic ProxyCommand support in #376.

NalinG:
Makes sense, thank you for the reply. We can probably close the PR.

PR #438, opened 2022-01-17
Add py.typed

q0w:
ref: https://www.python.org/dev/peps/pep-0561/#packaging-type-information


ronf:
Thank you! This is merged via commit d40c01b (with some minor edits).

PR #447, opened 2022-02-07
Increase minimum required fido2 version to 0.9.2

hexchain:
Related fido2 change: https://github.com/Yubico/python-fido2/commit/e9e56457d7f35cd3338e67573e6c43dc742b4b41


ronf:
Thanks very much for this! I noticed there was an issue with moving to fido 0.9.2, but hadn't had the chance to examine the history there to figure out what I needed to do to support it.

PR #455, opened 2022-02-20
Integrate with GitHub actions and codecov.io

hexchain:
This PR adds a GitHub actions workflow that runs the testsuite in all possible environments, compute the combined coverage, and (optionally) upload the coverage report to codecov.io.

Example workflow execution: https://github.com/hexchain/asyncssh/actions/runs/1871956917
Example codecov report: https://codecov.io/gh/hexchain/asyncssh/tree/90609cd4f3b98fe1f747c8cf6d63c90d6de1eae5

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,9 @@
 [run]
 branch = True
+relative_files = True
+source =
+    asyncssh
+    tests
 
 [report]
 exclude_lines =
@@ -9,6 +13,3 @@ exclude_lines =
 partial_branches =
     pragma: no branch
     for .*
-include =
-    asyncssh/*
-    tests/*
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -1,0 +1,104 @@
+name: Run tests
+on: [push, pull_request]
+
+jobs:
+  run-tests:
+    name: Run tests
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        openssl-version: ["default"]
+        include:
+          - os: macos-latest
+            python-version: "3.10"
+            openssl-version: "3"
+        exclude:
+          # test hangs on these combination
+          - os: windows-latest
+            python-version: "3.8"
+          - os: windows-latest
+            python-version: "3.9"
+          - os: windows-latest
+            python-version: "3.10"
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Determine pip cache path
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+      - name: Set up pip cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: pip-${{ runner.os }}-${{ hashFiles('setup.py', 'tox.ini') }}
+          restore-keys: |
+            pip-${{ runner.os }}-
+      - name: Determine vcpkg cache path
+        if: ${{ runner.os == 'Windows' }}
+        id: vcpkg-cache
+        run: |
+          echo "::set-output name=archives-dir::$env:LOCALAPPDATA\vcpkg\archives"
+          echo "::set-output name=downloads-dir::$env:VCPKG_INSTALLATION_ROOT\downloads"
+      - name: Set up vcpkg cache
+        uses: actions/cache@v2
+        if: ${{ runner.os == 'Windows' }}
+        with:
+          path: |
+            ${{ steps.vcpkg-cache.outputs.archives-dir }}
+            ${{ steps.vcpkg-cache.outputs.downloads-dir }}
+          key: vcpkg
+      - name: Install Linux dependencies
+        if: ${{ runner.os == 'Linux' }}
+        run: sudo apt install -y --no-install-recommends libnettle7 libsodium-dev libssl-dev libkrb5-dev ssh
+      - name: Install macOS dependencies
+        if: ${{ runner.os == 'macOS' }}
+        run: brew install nettle libsodium openssl
+      - name: Provide OpenSSL 3
+        if: ${{ matrix.openssl-version == '3' }}
+        run: echo "/usr/local/opt/openssl@3/bin" >> $GITHUB_PATH
+      - name: Install Windows dependencies
+        if: ${{ runner.os == 'Windows' }}
+        run: vcpkg install libsodium nettle openssl --triplet x64-windows
+      - name: Install Python dependencies
+        run: pip install tox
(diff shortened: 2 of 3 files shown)


ronf:
This looks very nice!

Is it possible to add in a new test run which installs OpenSSL 3 instead of OpenSSL 1.1.1 as a dependency? That'll fix the code coverage issue in test_public_key.py (line 86). I'm thinking perhaps we pick one version of Linux or macOS to run with the newer OpenSSL, or perhaps run the Python 3.10 test twice, once with OpenSSL 1.1.1 and once with OpenSSL 3.

I think most of the other reported code coverage issues are on Win32-specific files (agent_win32.py, gss_win32.py, and sspi_stub.py), and those are probably legitimate, as I never really focused on code coverage for those modules before now.

hexchain:
The macOS + Python 3.10 now runs exclusively with the OpenSSL 3.0 command line tool. The coverage on `test_public_key.py` is now 100%: https://app.codecov.io/gh/hexchain/asyncssh/blob/28130b1589b4f0a861d1c27af3238f6285f95d34/tests/test_public_key.py.

It turns out that there is a limit of 5 concurrent macOS runners for all GitHub plans other than Enterprise so adding another macOS runner will possibly double the total test time.

ronf:
Terrific - thanks very much!

Also, thanks for letting me know about the limit on concurrent runners. Using a mix of OpenSSL versions in the existing runners as you did seems like the best bet for now. At the point where it comes time to add a 3.11 test, I'll probably retire the 3.6 test, as 3.6 is technical end-of-life at this point (as of 23 Dec 2021). I won't go out of my way to intentionally break support for 3.6 any time soon, but testing that is less critical post-EOL.

Does that same limit apply to Windows and Linux as well?

hexchain:
> Does that same limit apply to Windows and Linux as well?

No, but there is a limit of total concurrent jobs: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

There are some other things to note, I'll put them in inline comments.

After merging this, you might want to check out the project page on codecov (https://app.codecov.io/gh/ronf/asyncssh/) to tweak some settings or set up integration with this GitHub repo.

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,108 @@
...
+        run: vcpkg install libsodium nettle openssl --triplet x64-windows
+      - name: Install Python dependencies
+        run: pip install tox
+      - name: Run tests
+        shell: python
+        run: |
+          import os, sys, platform
+          V = sys.version_info
+          p = platform.system().lower()
+          exit(os.system(f"tox -e py{V.major}{V.minor}-{p} -- -ra"))
```
hexchain, line 77: Here the toxenv name is generated from the Python version and the platform name. This is a bit complicated, but I cannot really think of a better way...

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,108 @@
...
+          name: coverage
+      - name: Install dependencies
+        run: |
+          sudo apt install -y sqlite3
+          pip install tox
+      - name: Report coverage
+        run: |
+          shopt -s nullglob
+          for f in .coverage.*-windows; do
+            sqlite3 "$f" "update file set path = replace(path, '\\', '/');"
```
hexchain, line 103: This is a hack to unify path separators so that coverage data can be combined.

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,108 @@
...
+        if: ${{ runner.os == 'Windows' }}
+        with:
+          path: |
+            ${{ steps.vcpkg-cache.outputs.archives-dir }}
+            ${{ steps.vcpkg-cache.outputs.downloads-dir }}
+          key: vcpkg
+      - name: Install Linux dependencies
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          sudo ip route add unreachable 0.0.0.1
```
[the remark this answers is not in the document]
ronf: I agree that this should be fixed in the tests. I actually made a change like this a couple of weeks ago (see commit bbb9fef), but hadn't gotten around to making similar changes in other places. I just checked that in as commit 5bb0b38, so you should be ok without this line now.

I experimented with a few different approaches for what to use for this, but eventually settled on an invalid address argument of '\xff', as that seemed to produce the right exception on all platforms without being affected by host network configuration. It also worked well for both connect() and listen(), which I needed in some places. The port 0 approach wasn't an option in that case.

.github/workflows/run_tests.yml
ronf: This looks ok, but does this mean you weren't able to make [paths] do this for you?

.coveragerc
```diff
@@ -9,6 +13,3 @@ exclude_lines =
 partial_branches =
     pragma: no branch
     for .*
-include =
-    asyncssh/*
-    tests/*
```
ronf, line 14: Was there a reason this needed to be removed? Without it, I sometimes see files from other directories popping up in the coverage output when I run things by hand.

hexchain: It was removed because in a prior test, this seems to prevent Windows coverage data from being combined due to the path separator. Also, this option only affects the reporting phase, and using similar option for the execution phase seems better.

Do files from unrelated directories still show up in tests? I have also added `[run] source` ([docs](https://coverage.readthedocs.io/en/latest/source.html#source)) to instruct coverage only measure the `asyncssh` and `tests` package. Any other packages should be ignored.

.github/workflows/run_tests.yml
hexchain: Unfortunately no. I'm also a bit hesitant to put any kind of absolute path in the configuration file, but it is what coverage wants. It also tries to make any non-absolute path absolute, which is why sometimes it complains about paths like `/home/runner/asyncssh/asyncssh/D:\a\asyncssh\asyncssh\asyncssh\__init__.py`.

.coveragerc
ronf: Ah - I missed that you moved these from [report] to [run]. I agree that's a better place.

The problem I saw turned out to be that I had copied the files to a new directory but didn't have the proper coveragerc in place.

Regarding the mixture of UNIX-style and Window-style paths, I've seen that in past experiments trying to merge coverage data between the two, but your last attempt at it here looked good to me.

Are there any remaining issues you're still working on it, or do you think this is good to go?

hexchain: I think it's good for now. I'm also happy to help with any related problems in the future.

ronf:
Thanks so much for this! It should now be merged, and I went over to codecov.io and set up a web hook for it.

I also see that a Github action has been queued, so we'll see how that goes...

ronf:
Everything looks good here - thanks again! This should be extremely useful, especially given that I lost my previous CI support for Linux and only had macOS locally.

Next step is to clean up some of the missing coverage on the SSPI stub on Windows, though at least some of those look like they may be a problem in the coverage tool. It's showing the body of various @property methods as being covered but the "def" line for those methods as not covered, which doesn't really make sense.

There's also some missing coverage on the Windows agent code, as I think right now I have agent unit tests disabled there. There is support for both ssh-agent on Windows 10 & later and for Pageant, but that would require those executables to be added and running in the test environment.

PR #458, opened 2022-03-02
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)


ronf:
Thanks for this! Based on my reading of https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls, I wonder if it might be worthwhile to also fill in a "Documentation" link as part of this, pointing at https://asyncssh.readthedocs.io. The "Requests" example you mentioned seems to have that in addition to the home page and link to source showing up on its page. AsyncSSH already has the home page link, but not the other two yet.

ronf:
Since I didn't hear back from you I went ahead and tried to push my own change which is a superset of what you proposed here. It's in commit 2fe1e6e, but I don't see the new links showing up on the PyPI page at https://pypi.org/project/asyncssh/. Perhaps there's a time lag before PyPI picks up this change? I see other edits I made in setup.py already reflected there, though - just not the project URLs.

ronf:
Looks like the changes in commit 2fe1e6e did eventually show up at https://pypi.org/project/asyncssh/ -- thanks again for suggesting this and providing an example!

PR #460, opened 2022-03-24
Account for size of payload length when sending data

### LINKED ISSUE #459: Data packets sent to dropbear can be oversized
Sending large amounts of data to a process's stdin through dropbear causes dropbear to complain with "bad packet, oversized decompressed" and close the channel.

Checked with asyncssh v2.9.0 (current master) and dropbear v2020.81. I see no related changes in asyncssh `develop` and on the dropbear side it seems like it should reproduce on all versions from v2013.59 to current master — anything since https://github.com/mkj/dropbear/commit/49263b5314e59de654d7b01f513b1cd95b1b9ec9 .

Reproducing this should amount to just having dropbear running somewhere — in case it makes a difference, I've seen this with `dropbear -W 65536` which increases the default receive window size from 24576 to 65536 — and then e.g.:

```python
import asyncio
import asyncssh
import logging

logging.basicConfig()
asyncssh.set_log_level(logging.DEBUG)
asyncssh.set_debug_level(2)

async def run():
    client = await asyncssh.connect(...)
    data = "x" * 50000
    result = await client.run("cat", input=data)
    assert result.stdout == data

asyncio.get_event_loop().run_until_complete(run())
```

To be precise, this works up to a payload size of 32758 but fails for 32759 or greater. This is obviously no coincidence, from an excerpt of the log output:

```
INFO:asyncssh:[conn=0, chan=0] Requesting new SSH session
DEBUG:asyncssh:[conn=0, chan=0]   Initial recv window 2097152, packet size 32768
DEBUG:asyncssh:[conn=0, chan=0]   Initial send window 65536, packet size 32759
INFO:asyncssh:[conn=0, chan=0]   Command: cat
DEBUG:asyncssh:[conn=0, chan=0] Sending 32759 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Sending EOF
DEBUG:asyncssh:[conn=0, chan=0] Reading from channel started
INFO:asyncssh:[conn=0] Connection failure: [Errno 104] Connection reset by peer
```

From what I can tell, dropbear negotiating a receive packet size of 32759 does not mean that it is willing to accept a `MSG_CHANNEL_DATA` with 32759 data bytes within it, but rather the whole `String(data)`, i.e. with the 4-byte length included, has to fit within those 32759 bytes, meaning there's room for only 32755 bytes in `data` itself. This feels somehow logical since it allows for 32768 bytes (dropbear's default `RECV_MAX_PAYLOAD_LEN`) to contain the 1-byte packet type, 4-byte channel, and in the case of `MSG_CHANNEL_EXTENDED_DATA` 4-byte data type, along with the rest of the payload. This snippet in dropbear's source suggests the same: https://github.com/mkj/dropbear/blob/a70b9e34c8119a82e05aaf4f7faa0ebf5a50dacc/common-channel.c#L62-L67

I don't know enough about the protocol to be able to say if this is actually an issue in asyncssh or in dropbear, but given the proliferation of old dropbear installations in embedded devices, it seems most practical to work around it in the client anyway.

Deewiant:
At least dropbear negotiates a precise packet size limit which must not
be exceeded by the sender, and that limit excludes only the 1-byte
packet type, 4-byte channel, and 4-byte data type (for
MSG_CHANNEL_EXTENDED_DATA). Thus, it includes in particular the 4-byte
payload length, which must also fit into the limit.

If the send window is exhausted the subtraction can cause the resulting
packet size to reach zero, which must also be handled to avoid sending
empty data packets.

Without this change, buffering more data than the negotiated packet size
will result in the sent packets triggering "bad packet, oversized
decompressed" errors from dropbear.

Fixes #459.

---

Especially having to do that `pktsize <= 0` check makes me feel like this probably isn't the appropriate fix, but at least I've tested that this works and I don't really have any alternative ideas.

To demonstrate that it's required at least in the context of this change, using e.g. a 100000-length payload in the test case in #459 shows the zero case triggering several times:

```
INFO:asyncssh:[conn=0, chan=0] Requesting new SSH session
DEBUG:asyncssh:[conn=0, chan=0]   Initial recv window 2097152, packet size 32768
DEBUG:asyncssh:[conn=0, chan=0]   Initial send window 65536, packet size 32759
INFO:asyncssh:[conn=0, chan=0]   Command: cat
DEBUG:asyncssh:[conn=0, chan=0] Sending 100000 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(65536, 32759) to 32755
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(32781, 32759) to 32755
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(26, 32759) to 22
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(4, 32759) to 0
DEBUG:asyncssh:[conn=0, chan=0] Sending EOF
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(4, 32759) to 0
DEBUG:asyncssh:[conn=0, chan=0] Reading from channel started
DEBUG:asyncssh:[conn=0, chan=0] Received window adjust of 32755 bytes, new window 32759
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(32759, 32759) to 32755
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(4, 32759) to 0
DEBUG:asyncssh:[conn=0, chan=0] Received window adjust of 32755 bytes, new window 32759
DEBUG:asyncssh:[conn=0, chan=0] Adjusted packet size from min(32759, 32759) to 32755
DEBUG:asyncssh:[conn=0, chan=0] Received window adjust of 34490 bytes, new window 65536
DEBUG:asyncssh:[conn=0, chan=0] Received 16375 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 16375 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 16375 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 16375 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 19 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 16375 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 16375 data bytes
DEBUG:asyncssh:[conn=0, chan=0] Received 1731 data bytes
INFO:asyncssh:[conn=0, chan=0] Received exit status 0
```

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -294,8 +294,13 @@ class SSHChannel(Generic[AnyStr], SSHPacketHandler):
         """Flush as much data in send buffer as the send window allows"""
 
         while self._send_buf and self._send_window:
-            pktsize = min(self._send_window, self._send_pktsize)
             buf, datatype = self._send_buf[0]
+            pktsize = min(self._send_window, self._send_pktsize)
+            pktsize -= 4  # 4-byte payload length
+            self.logger.debug2('Adjusted packet size from min(%d, %d) to %d',
+                               self._send_window, self._send_pktsize, pktsize)
+            if pktsize <= 0:
+                break
 
             if len(buf) > pktsize:
                 data = buf[:pktsize]


Deewiant:
Closing in favour of a better fix like https://github.com/ronf/asyncssh/issues/459#issuecomment-1082632195.

PR #464, opened 2022-03-28
Reintroduce backend arg to cipher.

sebby97:
I believe an incorrect change was made here asyncssh/crypto/chacha.py where a "backend" was removed and no longer passed into Cypher. The commit hash that includes the change is 88dcdac875e186bcd47b9c15f66b2935317033c6. The commit message mentions adding in TypeDefs so removing the arg was likely a mistake with the current minimum recommended version of cryptography==2.8. The change breaks the code path, however it is clear that future cryptograph versions intended to deprecate the backend arg, so the thing to do might be to unpin/update the minimum cryptography version.


ronf:
Closing, as this has been resolved - see the discussion in #465.

PR #480, opened 2022-05-18
Handle ConnectionRefusedError when connecting to 223.255.255.254

stefanor:
If the tests are run from an environment with a firewall, they may be refused instead of timing out.

Just skip the test.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -425,23 +425,32 @@ class _TestConnection(ServerTestCase):
     async def test_connect_timeout_exceeded(self):
         """Test connect timeout exceeded"""
 
-        with self.assertRaises(asyncio.TimeoutError):
-            await asyncssh.connect('223.255.255.254', connect_timeout=1)
+        try:
+            with self.assertRaises(asyncio.TimeoutError):
+                await asyncssh.connect('223.255.255.254', connect_timeout=1)
+        except ConnectionRefusedError:
+            raise unittest.SkipTest('Outbound connection firewalled')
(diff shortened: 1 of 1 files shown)


ronf:
Thank you for this. I wish there was a better way to test with an address that will reliably get no immediate response from the stack. I was originally using various flavors of 0.* and loopback addresses, different OSes handled these differently and so I keep looking for better options. Perhaps I should just resort to a mock here, and force the connect timeout without relying on the stack. However, until I get to that, this seems like a reasonable change.

Minor nit: Could you switch over your string constants to use single quotes rather than double quotes, for consistency with the rest of the code?

stefanor:
Yeah, I can't think of any way to reliably get that, short of patching.

> Minor nit: Could you switch over your string constants to use single quotes rather than double quotes, for consistency with the rest of the code?

Done! Sorry about the delay.

ronf:
Thanks again for your submission here. It occurred to me that getting code coverage on this would be a bit messy, though, as not all environments would let me trigger the new exception block you added. In the end, it was actually pretty simple to do the mock instead, and that avoids other potential errors from triggering depending on the local networking environment.

I've checked in an alternative fix for this in commit e59262f in the "develop" branch that I hope will address the underlying issue that you ran into.

I really appreciate your report! If you're able to test this alternate fix in your environment, I'd really appreciate it. If you have any problems with it, please let me know...

ronf:
Closing this given the alternate fix, but thanks again for raising this!

PR #481, opened 2022-05-21
add 'pattern' attribute to SFTPGlob.match() exception

LuckyDams:
Allow new `SFTPGlob.match()` to return pattern value when raising `OSError` or `SFTPError` exceptions. This would allow to define which pattern is returning 'No matches found' error when using more than one patterns in `SFTPClient.glob()` and `SFTPClient.glob_sftpname()` methods.

Please note, in previous implementation `SFTPClient.glob()` was returning pattern value in 'srcpath' attribute which is now returning plain path portion of a pattern.

Proposed solution here is to (simply) add new 'pattern' attribute to exception.

### DIFF
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -2152,6 +2152,7 @@ class SFTPGlob:
                 self._report_match(path, attrs)
         except (OSError, SFTPError) as exc:
             setattr(exc, 'srcpath', path)
+            setattr(exc, 'pattern', pattern)
 
             if error_handler:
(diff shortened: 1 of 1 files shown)


ronf:
Actually, I didn't mean to change this. I'll set it "srcpath" back to the pattern as it was before, preserving the old behavior.

ronf:
Fixed in 8e2192e - thanks for catching this!

LuckyDams:
Welcome. Was unsure of the intent here.

PR #484, opened 2022-06-05
Close send channel on channel close

tahajahangir:
### Short Description
After the client closed the connection, the calls to `stdout.write` should be interrupted with `BrokenPipeError`, so the `_send_state` should be set to `closed`.

### Long Description

Run the server in *Server Examples > Line Editing*
https://asyncssh.readthedocs.io/en/stable/#line-editing

Connect with a ssh-client, and close the connection (e.g with `\n~.`) while the server is wait for input. The `handle_client` coroutine  continues to handle stdin/stdout streams and exits without any Exception (without noticing of a closed connection).

### DIFF
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -233,7 +233,7 @@ class SSHChannel(Generic[AnyStr], SSHPacketHandler):
 
             assert self._recv_chan is not None
             self._conn.remove_channel(self._recv_chan)
-            self._send_chan = None
+            self._close_send()
             self._recv_chan = None
             self._conn = None
 


ronf:
Thanks for the report!

As the code currently stands, the handle_client() function is actually receiving EOF in the form of an empty string returned from the `read(1)` call. If this example involved a read loop or using stdin as an iterator, that EOF would generally be enough to cause the handler to exit. So, I don't think there's a problem with the handler continuing to handle stdout/stderr after the client closes. The example code doesn't exit because it's waiting for new connections, but the handler for that particular connection DOES properly exit as soon as the client connection close is received.

That said, I agree with you that the call to `write('\n')` should have triggered a `BrokenPipeError` since it is called AFTER the application has been notified of the connection close.

In terms of the fix, the `_cleanup()` functions in AsyncSSH are designed to be called as the very last thing  to finish locally cleaning up their object, so I don't think it's a good idea to have `_cleanup()` call `_close_send()`, since in some cases that calls `send_packet(MSG_CHANNEL_CLOSE)`. Also, in this case in particular, the cleanup is triggered by a call to `process_connection_close()`, which means the code knows the SSH connection has already been closed by the peer, so there's really no point in sending close messages back for any remaining open channels. However, I agree that `process_connection_close()` does need to trigger `_send_state` being set to `'closed'`, and it's probably also a good idea to explicitly discard the send buffer.

I'm leaning toward the following fix:

```patch
diff --git a/asyncssh/channel.py b/asyncssh/channel.py
index a957138..c4440a7 100644
--- a/asyncssh/channel.py
+++ b/asyncssh/channel.py
@@ -453,6 +453,8 @@ class SSHChannel(Generic[AnyStr], SSHPacketHandler):

         self.logger.info('Closing channel due to connection close')

+        self._send_state = 'closed'
+        self._close_send()
         self._cleanup(exc)

     def process_open(self, send_chan: int, send_window: int, send_pktsize: int,
```
Forcing the state to `'closed'` before calling `_close_send()` will make sure that no channel close messages are sent in this case, while still allowing this to share the code to discard the send buffer contents. Other calls to `_close_send()` will also continue to be able to send channel closed messages when appropriate.

Looking more closely at the code, `_cleanup()` for channel objects is currently ONLY called from `_process_connection_close()`, so another option would have been to put this code in `_cleanup()` itself similar to your change (but still forcing the state to `'closed'` first). However, if there's a need at some point to call `_cleanup()` from somewhere else, I think it's best to make sure that will never indirectly trigger a call to `send_packet()`.

Issue #485, opened 2022-06-07
Pyright doesn't know exported symbols


nagua:
Hello again,

while trying to code an application using your library, I noticed that pyright/pylance doesn't consider the symbols in `asyncssh.*` to be exported. You have a `py.typed` file that buys into a special behavior. Every symbol imported in the `__init__.py` is treated as an private import, unless you use a `redundant module import`. You can read about it here: https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface the second bullet point in the second list is the relevant part to read.

It would be really nice and easier to use, if this could be fixed. If I have enough time, I will try to build a pull request for that!

Have a nice day!

nagua:
I tried to run `mypy` now but it shows me 74 errors. Is there something specific I need to do to run `mypy`?
Also I just changed all exports in the `asyncssh/__init__.py` to `from import as` style and `pyright` is happy for now.
(At least for autocompletion). So I'm currently not sure about the incompatibilities that you mentioned. If you want to,
I can make a PR with my changes so you can decide if this is a problem or not.

ronf:
Which version of mypy are you running?

With mypy 0.961 on Python 3.10, I'm seeing a small number of errors here that were masked by some local stubs I've been running with to get better type coverage for modules that were missing annotations. However, even running without those stubs, I get only 5 errors which look like they're all straightforward to correct. I'll do that.

As for pyright, changing every single one of the hundreds of symbols in __init__.py to match the "from x import y as y" syntax didn't really appeal to me, especially considering that Python itself is clearly treating these symbols as public in the "asyncssh" module without requiring that. I don't see why "pyright" shouldn't do the same.

The other option appears to be to list all the symbols in an __all__ definition in __init__.py. That's slightly less ugly to read, but it still ends up duplicating the names of all of these symbols twice in the module, which I'd really like to avoid.

I'm also concerned that pyright run on application code using AsyncSSH might produce a significant number of errors similar to those I saw when attempt to run pyright on the AsyncSSH code itself. In a quick test, the latest version of pyright (1.1.253) on Python 3.10 returned 536 errors and 1 warning despite mypy running cleanly. Some of these are potentially legitimate errors caught by pyright that were missed by mypy, but even if that's the case I'm not sure I'll have the cycles available any time soon to go through the complete list and figure that out. Also, when I last attempted this, there were many errors I couldn't "correct" without major rewrites of working code, risking breaking things in the process.

[...]

PR #489, opened 2022-06-20
Add __all__ dictionary, for pyright completion

nagua:
Hello @ronf,

as discussed in #485 I created added a `__all__` symbol to the `__init__.py` file.

What do you think of this?

Have a nice day!


ronf:
Thanks for this. I'm probably going to want to trim this a little so that it only lists symbols which are meant to be public. In particular, the packages at the end (lines 265-272) should not be considered public, even though I had to import them to trigger register function calls in them. I'll also probably want to reorder the list to be alphabetical to make it easier to see whether a symbol is present or not, and also put more than one entry per line so more of the symbols can be visible at once.

That said, the important question is whether this was sufficient to get pyright to work the way you wanted here. Was it?

nagua:
Yes, `pyright` can now successfully give autocompletions for `asyncssh`. I can sort the list if you want. I didn't want to have all modules in one line and I couldn't find any formatting guidelines, so I just used `black` for formatting and that was what it came up with. So feel free to suggest a different formatting scheme.

For the private modules: Should I just delete lines 265-272?

ronf:
> For the private modules: Should I just delete lines 265-272?

Yes - I think that'd probably be best.

There are a handful other symbols in the list which are currently undocumented and not really meant for client use, but some of them show up in type annotations, so perhaps they should be included. I'm thinking of things like SSHForwarder and various  "Factory" types.

Regarding the formatting, I generally try to line wrap somewhere around 75 characters, occasionally letting it go to 79
if needed to break on a word boundary. Anything wider should wrap a word earlier, even if that makes the line  less than 75 characters wide, but fitting as much as it can on each line within those limits. I'm good with the `__all__ = [` and the trailing `]` on lines by themselves, and everything in between being indented 4 spaces. It'd just be good to sort the symbols and then combine them onto fewer lines.

Also, it looks like a space got added on line 76. It might be good to remove that to reduce noise.

llchan:
Might I suggest that we sort the `__all__` list so that it's simpler to merge/insert lines?

ronf:
Yes, that's the plan.

ronf:
Thanks, @nagua !

I was able to confirm here that this change does silence the pyright warnings, and the formatting generally looks good. Just one final request before I merge it: Could you change all the double quotes to single quotes? The AsyncSSH default style is to use single quotes for all string constants, unless a string contains single quotes in its value.

nagua:
Ok, the quotes are single quotes now.

ronf:
Thanks so much, @nagua -- this is now merged into the "develop" branch and will be included in the next release!

PR #495, opened 2022-07-09
Guard win32 imports on non-win32 systems

gsauthof:
This fixes a unit test case error on Fedora Rawhide.

See also: https://kojipkgs.fedoraproject.org//work/tasks/5653/89275653/build.log

It's basically a roll-back of: https://github.com/ronf/asyncssh/commit/a89241a778b09e117cd5673acbd3d7bc38b90468

Perhaps I'm missing something - but I don't think that (even if one would install the sspi win32 wrapper package under Linux) this could increase the coverage under Linux because the stub usage is also guarded:

https://github.com/ronf/asyncssh/blob/161ba01bcc5f7fa207e12c9b85adb5d1dd9ffc66/tests/util.py#L120-L125

### DIFF
--- a/tests/sspi_stub.py
+++ b/tests/sspi_stub.py
@@ -21,8 +21,12 @@
 
 """Stub SSPI module for unit tests"""
 
-from asyncssh.gss_win32 import ASC_RET_INTEGRITY, ISC_RET_INTEGRITY
-from asyncssh.gss_win32 import SECPKG_ATTR_NATIVE_NAMES, SSPIError
+
+import sys
+
+if sys.platform == 'win32': # pragma: no cover
+    from asyncssh.gss_win32 import ASC_RET_INTEGRITY, ISC_RET_INTEGRITY
+    from asyncssh.gss_win32 import SECPKG_ATTR_NATIVE_NAMES, SSPIError
 
 from .gss_stub import step
 


ronf:
The previous change was meant to just get coverage to 100% in this specific module. As you noted above, the usage of this was separately covered with a "no cover" pragma.

I'm not sure why I didn't see the build failure you did on macOS or Linux, where "sspi" is not installed. However, I'd like to keep the coverage here if possible. It looks like it should be sufficient to add a "# pragma: no cover" on the platform conditional you added back, similar to what it present in util.py. Could you make that change?

In theory, this shouldn't be needed if coverage measurements are merged for all platforms, but my local test setup is still per-OS, so it has been useful for now to exclude the win32-only blocks for now.

gsauthof:
Sure, I've added the pragma.

Is it possible that you had the sspi package installed on non-win32 environments as well, by chance?

I haven't tried it but it looks like it's Python only and it's even packaged for Fedora (as a dependency of pypykatz - a Python implementation of Mimikatz, of all things ...).

ronf:
Thanks for the updated PR! I'll merge it in shortly.

I did check on whether "sspi" was installed or not in my local environment, and I confirmed prior to my last reply that "sspi" is not installed in any of my local Python installs on either macOS or Linux. It looks like it is located in a "win32" directory and depends on other win32 packages (sspicon, win32security), so it's not the kind of thing that could be easily installed elsewhere without repackaging.

It's interesting that Fedora actually has its own port of that, but I don't see it on PyPI (though there are other lesser known pure Python SSPI implementations on there).

gsauthof:
FWIW, Fedora has this: https://github.com/skelsec/winsspi (https://pypi.org/project/winsspi/)

ronf:
Ah, ok - that's one of the "lesser known" versions I mentioned.

PR #509, opened 2022-09-13
Fix "Recceived window change" in log message

telnet23:
Remove the extra "c"


ronf:
Thanks very much for catching this!

Issue #511, opened 2022-09-15
Server fails to open remote tunnel with dynamic port


daniel-shimon:
Hi!
Running an asyncssh server with tunnels enabled, I'm trying to dynamically allocate a listen port (using a normal openssh client), aka `ssh -R 0:DEST_HOST:DEST_PORT`.
The tunnel seems to be created fine, but when trying to use it the client shows the following error:
```
Allocated port 56704 for remote forward to 127.0.0.1:8022
WARNING: Server requests forwarding for unknown listen_port 0
```

After digging in and running with `set_debug_level(3)` I can see the issue:
```
### In tunnel creation:
[conn=0] Creating local TCP forwarder on localhost
[conn=0] Assigning dynamic port 56704
[conn=0] Created TCP listener on localhost, port 56704

...
### In tunnel usage:
[conn=0] Opening forwarded TCP connection to localhost
[conn=0]   Client address: ::1, port 56721
[conn=0, chan=0] Set write buffer limits: low-water=16384, high-water=65536
[conn=0, chan=0]   Initial recv window 2097152, packet size 32768
[conn=0, pktid=14] Sent MSG_IGNORE (2), 5 bytes
  00000000: 02 00 00 00 00                                   .....
[conn=0, chan=0, pktid=15] Sent MSG_CHANNEL_OPEN (90), 60 bytes
  00000000: 5a 00 00 00 0f 66 6f 72 77 61 72 64 65 64 2d 74  Z....forwarded-t
  00000010: 63 70 69 70 00 00 00 00 00 20 00 00 00 00 80 00  cpip..... ......
  00000020: 00 00 00 09 6c 6f 63 61 6c 68 6f 73 74 00 00 00  ....localhost...
  00000030: 00 00 00 00 03 3a 3a 31 00 00 dd 91              .....::1....
[conn=0, pktid=8] Received MSG_CHANNEL_OPEN_FAILURE (92), 28 bytes
  00000000: 5c 00 00 00 00 00 00 00 01 00 00 00 0b 6f 70 65  \............ope
  00000010: 6e 20 66 61 69 6c 65 64 00 00 00 00              n failed....
[conn=0, chan=0] Channel closed
```

In the `MSG_CHANNEL_OPEN` message, the source host is '0.0.0.0' but the host port is also 0 instead of the allocated port 41547.
Because of this the client doesn't know to which tunnel the packet belongs to and returns a `MSG_CHANNEL_OPEN_FAILURE` message.

[...]

PR #512, opened 2022-09-15
Draft: Fix forwarding a dynamically allocated port in the server

daniel-shimon:
In case a dynamic port was requested (i.e. `listen_port == 0`), the `listen_port` variable is later set to the allocated port

This fixes https://github.com/ronf/asyncssh/issues/511


daniel-shimon:
This quick fix works for me but probably isn't the correct fix to do, could I get some help doing it correctly in the project?
Thanks!

PR #518, opened 2022-10-17
Fixed endless read with exact=True.

KoBOLL:
Assume that read called as read(10, None, True).
Assume that 5 bytes is already read and TerminalSizeChanged raised. After checking that recv_buf[0] is instance of Exception and some data is received we should break **outer** while loop to get IncompleteReadError raised.

P.S.
I'm sorry I don't understand how to indicate that I OK with EPL 2.0 before pull request.
So I'm OK with EPL 2.0 :)

### DIFF
--- a/asyncssh/stream.py
+++ b/asyncssh/stream.py
@@ -511,10 +511,12 @@ class SSHStreamSession(Generic[AnyStr]):
         data: List[AnyStr] = []
 
         async with self._read_locks[datatype]:
+            break_read = False
             while True:
                 while recv_buf and n != 0:
                     if isinstance(recv_buf[0], Exception):
                         if data:
+                            break_read = True
                             break
                         else:
                             exc = cast(Exception, recv_buf.pop(0))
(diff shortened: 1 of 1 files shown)


ronf:
Thanks for the report, and a proposed fix!

Given the way exceptions are returned inline, I agree that on a readexactly() call, raising an IncompleteReadError is probably the best that can be done, even though the stream is not yet at EOF. If a caller wanted to, they could buffer the partial data when this happened, read any exceptions, and then read the remainder of the input with another (shorter) readexactly() if that turned out to be something they needed to support.

Note that there may also be an issue here around a Unicode character being split before and after the exception on a stream with encoding set. However, in that case, I think the Unicode character won't actually be added to the receive buffer until it is complete. So, effectively these partial bytes will only be added to the receive buffer _after_ a signal arrives even if partial data for that character arrived before. At the time of the signal, this data will still be buffered in an IncrementalDecoder.

I'm going to take a look at adding some explicit unit tests around this case to help confirm the fix and avoid regressions later, and then I'll look into merging this in.

ronf:
Thanks very much for this! I moved this change over to the "develop" branch and added a unit test to validate the fix. The result is available as commit cbb485c. This change will be included in the next release.

ronf:
This fix is now available in AsyncSSH 2.13.0.

PR #519, opened 2022-10-18
Fix deprecation notices in workflows

hexchain:
Fixes most of the deprecation notices.

There is still one left about `set-output` inside `actions/download-artifact@v3`, which we can only wait for the upstream to fix: https://github.com/actions/download-artifact/issues/185.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -23,34 +23,40 @@ jobs:
             python-version: "3.10"
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


ronf:
Thanks very much!

ronf:
@hexchain I ran into an issue today with ubuntu-latest where I had to update the libnettle dependency and remove Python 3.6 support. I'm guessing it has to do with ubuntu-latest moving to be Ubuntu 22.04. These changes seem to be working, as does a change I made to add Python 3.11. However, the 3.11 builds aren't generating coverage files for some reason. Other versions don't show this warning. Any idea what might be happening?

hexchain:
I guess the coverage file name could have been changed. Will take a look ASAP.

hexchain:
Please see #531.

ronf:
Thanks very much!

PR #531, opened 2022-12-18
Update workflow and tox.ini for Python 3.11 and liboqs

hexchain:


.github/workflows/run_tests.yml
```diff
@@ -100,7 +100,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.7"
+          python-version: "3.11"
```
ronf, line 103: There's a bug in coverage which is triggered when coverage data is captured on 3.7 or earlier but the report is generated on 3.8 or later. See e9f0e5d5 for the previous change which fixed this. I think this change needs to be reverted to maintain coverage reporting on Windows, which is currently only running tests on Python 3.7 due to other issues.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.7
-envlist = clean,{py36,py37,py38,py39,py310}-{linux,darwin,windows},report
+envlist = clean,{py37,py38,py39,py310,py311}-{linux,darwin,windows},report
```
ronf, line 3: Aha - I completely missed updating tox.ini for 3.11... Thanks for catching this!

.github/workflows/run_tests.yml
hexchain: Reverted

tox.ini
hexchain: On second thought this is probably not needed anymore, let me try (on another branch) to remove this and see what happens...

hexchain: Okay, now `tox.ini` no longer contains hardcoded Python version numbers. The only downside is, if you have multiple Python installed and want to run a test across all of them by simply executing `tox`, it wouldn't work anymore.

ronf: I actually do make use of the envlist to run "tox -p" locally, so I think I'd like to keep that. Is there any way to leave that in but avoid having to repeat things down in the "depends"? If not, I'm ok with just taking the COVERAGE_FILE changes.

On another note, there's a new dependency on "liboqs" (for post-Quantum encryption) that we might want to try and add here. On macOS, it's available via HomeBrew, so that should be pretty easy. Perhaps that'd be good enough for now to make codecov happy. Unfortunately, I don't see a precompiled version of liboqs in Ubuntu, so it would probably need to be compiled from source for Ubuntu and Windows. More info is at https://github.com/open-quantum-safe/liboqs.

hexchain: Noted, I'll restore the envlist and try to reduce some repetition if possible. I'll also look into building and installing liboqs.

ronf: Thanks! I made the change to add liboqs for macOS already, and it seemed to work nicely.

.github/workflows/run_tests.yml
```diff
@@ -63,32 +84,65 @@ jobs:
...
+          sudo cmake --install build
+
+      - name: Install liboqs (Windows)
+        if: ${{ runner.os == 'Windows' }}
+        shell: pwsh
+        working-directory: liboqs
+        run: |
+          cmake -GNinja -Bbuild . -DBUILD_SHARED_LIBS=ON -DOQS_BUILD_ONLY_LIB=ON -DCMAKE_C_COMPILER_LAUNCHER=sccache
+          cmake --build build
+          copy build\bin\liboqs.dll build\bin\oqs.dll
```
hexchain, line 126: For some reason, the built DLL is called `liboqs.dll` which cannot be found by Python with `find_library("oqs")`.

tox.ini
```diff
@@ -23,27 +26,33 @@ platform =
...
 parallel_show_output = true
+setenv =
+    COVERAGE_FILE =
 commands =
     coverage combine
     coverage report --show-missing
     coverage html
     coverage xml
+depends =
+    py3{7,8,9,10,11}-{linux,darwin,windows}
```
hexchain, line 55: Unfortunately, the repetition seems unavoidable, but at least it can be copied/pasted as a whole line now.

ronf: No problem - thanks for looking into this. This is definitely an improvement.

.github/workflows/run_tests.yml
ronf: I've had trouble with this in the past. For nettle, I ended up with the following in `crypto/umac.py`.

```python
_nettle_lib = 'libnettle-6' if sys.platform == 'win32' \
                            else ctypes.util.find_library('nettle')

if _nettle_lib: # pragma: no branch
    _nettle = ctypes.cdll.LoadLibrary(_nettle_lib)
```

In this case, the DLL was libnettle-6.dll, installed in the Python version's main directory (`C:\Program Files\Python310` on my system). I think if you make a similar win32 check and skip find_library, something similar to the above should hopefully work with liboqs.

Since I haven't tried building liboqs on Windows yet, I didn't make the same change there yet.

hexchain: I think I figured out why. If CMake uses the GNU toolchain it will add the `lib` prefix to the DLL. I've included a third-party action that initializes an MSVC environment so that CMake will use the MSVC toolchain.

hexchain: Not sure if it's the same reason for nettle, though. It seems to be a lot trickier when vcpkg is involved.

ronf: So find_library() works, but only when the filename is oqs.dll, not liboqs.dll? I'm not a big windows user, but that wouldn't surprise me. What's strange, though, is that other third-party packages I use seem to want the "lib" version. For instance, pynacl seems to look for "libsodium-<version>":

```python
__SONAMES = (23, 18, 17, 13, 10, 5, 4)

def _get_nacl():
    '''
    Locate the nacl c libs to use
    '''
    if sys.platform.startswith('win'):
        try:
            return ctypes.cdll.LoadLibrary('libsodium')
        except OSError:
            pass
        for soname_ver in __SONAMES:
            try:
                return ctypes.cdll.LoadLibrary(
                    'libsodium-{0}'.format(soname_ver)
                )
            except OSError:
                pass
        msg = 'Could not locate nacl lib, searched for libsodium'
        raise OSError(msg)
```

For Nettle. the binaries I found online has a similar naming convention (libnettle-6.dll), but interestingly it seems that at least on win64 they actually went with the bane 'hogweed.dll' without a prefix/suffix for another library nettle depended on!

Knowing that there is a way to make this work with find_library() on Windows, perhaps I could look into another approach where I try find_library() first, and only when it doesn't find it do I have it fall back to the "lib" version of the name. For instance:

```python
_nettle_lib = ctypes.util.find_library('nettle') or 'libnettle-6'
```

Actually, playing around a bit, I can actually pass in something like 'libnettle-6' directly to find_library() and it'll return a full path for the DLL, so I could also just have it try 'nettle', 'libnettle', and 'libnettle-6' until it found one that works.

For the sake of your check-in here, I'm good with just leaving it as the existing find_library() call if you have that working. I'll think about f I want to allow for the "lib" version of the name as a fallback, but that can be done as a separate checkin.

Thanks for your work on this! Let me know when you've got something you're comfortable merging.

hexchain: > So find_library() works, but only when the filename is oqs.dll, not liboqs.dll? I'm not a big windows user, but that wouldn't surprise me.

Yes, it seems to be the case. On Linux/Unix `find_library` expects the library name without `lib`, on Windows, it is just the name of the DLL.

>What's strange, though, is that other third-party packages I use seem to want the "lib" version. For instance, pynacl seems to look for "libsodium-":
> ...
> 
> For Nettle. the binaries I found online has a similar naming convention (libnettle-6.dll), but interestingly it seems that at least on win64 they actually went with the bane 'hogweed.dll' without a prefix/suffix for another library nettle depended on!
> 

My understanding is that the naming of the DLL depends on the build system, which usually makes its decision based on the discovered toolchain. In both CMake and Meson, DLLs have the `lib` prefix when they are built with the MinGW toolchain, but not with MSVC. Specifically, CMake automatically sets `CMAKE_SHARED_LIBRARY_PREFIX` to `lib` if the toolchain is GNU (MinGW), or keeps it empty if it detects MSVC.

`hogweed.dll` remains a mystery for me - it is a part of Nettle, and their Autotools build system doesn't seem to support MSVC, and in its [configure.ac](https://github.com/gnutls/nettle/blob/master/configure.ac) it refers to the library as `libhogweed.{so,dylib,a,dll,...}`, so I'm not sure how that file is produced...

> Knowing that there is a way to make this work with find_library() on Windows, perhaps I could look into another approach where I try find_library() first, and only when it doesn't find it do I have it fall back to the "lib" version of the name. For instance:
> 
> ```python
> _nettle_lib = ctypes.util.find_library('nettle') or 'libnettle-6'
> ```
> 
> Actually, playing around a bit, I can actually pass in something like 'libnettle-6' directly to find_library() and it'll return a full path for the DLL, so I could also just have it try 'nettle', 'libnettle', and 'libnettle-6' until it found one that works.
> 

That sounds like a good idea!

hexchain:
I think it's good to go now.

Note that if you find the Windows job constantly taking more than 30 minutes, you could try to clear the "vcpkg" cache at https://github.com/ronf/asyncssh/actions/caches.

.github/workflows/run_tests.yml
ronf: Ok - I've gone ahead and made the change mentioned above around load_library() for Nettle and OQS in commit f1f36ce. For OQS in particular, it should support either "oqs.dll" or "liboqs.dll" on Windows.

I'll be merging this PR shortly - I really appreciate your help on this!

hexchain: No problem, happy to help!

ronf:
Regarding clearing vcpkg from cache, I gave that a try just before check-in on the library naming.. We'll see if that helps speed things up. Thanks for the tip -- about 31 mins out of 45 mins total on the Windows runs seems to be on setting up dependencies.

ronf:
Minor improvement on Windows build time (45m -> 37m) after clearing vcpkg from the cache, but it looks like the run following that is back up to 43m, once again spending 31m on setting up Windows dependencies.

hexchain:
Hmm, I'll submit a PR later to optimize this vcpkg cache (along with some other small improvements I just discovered).

PR #532, opened 2022-12-24
CI optimizations

hexchain:
Some optimizations to the GH Actions workflow.

- The `pip` cache is now handled by `actions/setup-python`. Our custom cache code is thus removed.
- Vcpkg is no longer used and [an unofficial prebuilt version of libnettle](https://github.com/ShiftMediaProject/nettle) is downloaded each time instead.

  I just realized that:
  1. vcpkg-installed packages are not automatically made available to the system
  2. its nettle package does not produce a dll at all when building with MSVC.
 
  That means code depending on nettle was never really tested. This is now fixed, and `asyncssh/crypto/umac.py` is now fully tested on Windows.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -48,19 +48,10 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-
-      - name: Determine pip cache path
-        shell: bash
-        run: |
-          echo "_PIP_CACHE_DIR=$(pip cache dir)" >> $GITHUB_ENV
-
-      - name: Set up pip cache
-        uses: actions/cache@v3
-        with:
-          path: ${{ env._PIP_CACHE_DIR }}
-          key: pip-${{ runner.os }}-${{ hashFiles('asyncssh/setup.py', 'asyncssh/tox.ini') }}
-          restore-keys: |
-            pip-${{ runner.os }}-
+          cache: pip
+          cache-dependency-path: |
+            asyncssh/setup.py
+            asyncssh/tox.ini
 
       - name: Determine vcpkg cache path
         if: ${{ runner.os == 'Windows' }}
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -30,6 +30,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       liboqs_version: '0.7.2'
+      nettle_version: nettle_3.8.1_release_20220727
 
     steps:
       - name: Checkout asyncssh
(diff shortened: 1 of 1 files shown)


ronf:
Wow - this sounds great, and the workflow file looks a lot simpler!

Regarding nettle and Windows, I had previously been using Appveyor for my Windows builds, and I did something similar about using precompiled binaries. I hosted them myself, though, so I didn't need to work about a third-party link breaking, though that had a different problem in my never updating the version of those files. They're all still from 2016/2017, probably around when I first got Appveyor set up.

I'm happy to go with the external link you found, but if there is a way to get a compile from source to work, that'd probably be preferable. What did you end up doing for libsodium? I would think it would have a similar issue, though I did notice that libnacl might now be bundling a compiled version of that along with the Python wrapper. When I first started using libnacl, I don't think that was the case, and you had to install libsodium manually in order for libnacl to work.

hexchain:
For nettle, I think the current situation is fine for the moment, because
1. The project (ShiftMediaProject/nettle) has been around for a long time (6 years) and is doing a decent job at porting nettle to MSVC. IMO this is more "native" and robust than simply building it with MinGW. This is just a feeling though, I don't know Windows enough to judge how a MinGW-compiled library compares to an MSVC-compiled one, especially when LoadLibaray is involved.
2. I haven't been able to figure out how to reliably build nettle with vcpkg. The [vcpkg port of nettle](https://github.com/microsoft/vcpkg/tree/master/ports/nettle) does not produce a DLL with x64-windows, and fails when linking with x64-mingw-dynamic. Also, I am trying to avoid a pure source build (like the current liboqs) because it is hard to cache correctly.

For libsodium, there is no binary in libnacl so it won't work if there isn't libsodium in the system. But it seems that related routines only get a chance to run if PyCA cryptography fails to provide some features. This is hardly the case since the cryptography wheel contains platform binary. Also, most of these routines are currently excluded from coverage. How do you usually test those code paths? Maybe I can also make it work in CI.

ronf:
Regarding libnacl not including libsodium, that was my experience as well, but I thought maybe that had changed recently. Looking more closely, it turns out that what I saw was a different wrapper called PyNaCl (https://pypi.org/project/PyNaCl/), which can use either a bundled version of libsodium or be told to use a version already installed on the system. AsyncSSH is still using libnacl, though, so testing that would require having libsodium to be separately built/installed.

That said, you're right that there isn't really much need for libnacl any more. I originally used it for curve25519, ed25519,  and chacha20-poly1305 before the cryptography package had support for these. The libnacl version is now only used as a fallback, if cryptography (or the version of OpenSSL it is linked against) is missing support for any of these.

Since the code using libnacl is basically frozen at this point, I haven't worried about actively testing it or measuring coverage on it. That code has no dependencies on other parts of AsyncSSH I might be changing, so other changes shouldn't break it. In theory, it could be tested by mocking out the calls to things like backend.poly1305_supported() in cryptography, but it would have to be done before the associated crypto module is loaded. If there was a problem with this code in the future, I'd be more likely to just remove that code and fully depend on the version in cryptography/OpenSSL at this point.

PR #551, opened 2023-03-26
Link to PyPI and documentation

gsauthof:
Link prominently to the PyPI and documentation packages from the README.

The PyPI link is important to establish as clear as possible what the official PyPI project name is. For example, for new users who are quick to pip install it, they can then simply follow that link and copy and paste the PyPI name to avoid typos on the command line (to protect against typosquatting). Also there are projects where github reponame is unequal to the PyPI name, thus, such a link clears it up.

Having the documentation link placed on top helps to getting started, clarifies that this package indeed has an extensive documentation and is the most likely and fastest way to the changelog.


ronf:
Thanks for this!

PR #569, opened 2023-05-28
Avoid use of datetime.utcfromtimestamp, deprecated in Python 3.12

scop:


ronf:
Thanks for this change!

PR #578, opened 2023-07-11
Read from stdout instead of communicate

gsauthof:
This fixes the failure of the test_stdout_stream case when running under Python 3.12 (beta), on Fedora Rawhide (i.e. pre-F39).

When using communicate() there is a race where stdin is closed too early.

---

Failure log:  https://kojipkgs.fedoraproject.org//work/tasks/9302/103229302/build.log

```
======================================================================
FAIL: test_stdout_stream (tests.test_process._TestProcessRedirection.test_stdout_stream)
Test with stdout redirected to an asyncio stream
----------------------------------------------------------------------
  File "/builddir/build/BUILD/asyncssh-2.13.2/tests/util.py", line 97, in async_wrapper
    return self.loop.run_until_complete(coro(self, *args, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 664, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/builddir/build/BUILD/asyncssh-2.13.2/tests/test_process.py", line 907, in test_stdout_stream
    self.assertEqual(stdout_data, data.encode('ascii'))
AssertionError: b'' != b'4121131720'
----------------------------------------------------------------------
Ran 1504 tests in 2216.356s
FAILED (failures=1, skipped=53)
```

Success log (with the fix):  https://kojipkgs.fedoraproject.org//work/tasks/3111/103233111/build.log

```
test_stdout_stream (tests.test_process._TestProcessRedirection.test_stdout_stream)
Test with stdout redirected to an asyncio stream ... ok
[..]
Ran 1504 tests in 713.627s
OK (skipped=53)
```

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -902,7 +902,7 @@ class _TestProcessRedirection(_TestProcess):
             proc1.stdin.write(data)
             proc1.stdin.write_eof()
 
-            stdout_data, _ = await proc2.communicate()
+            stdout_data = await proc2.stdout.read()
 
         self.assertEqual(stdout_data, data.encode('ascii'))
 


gsauthof:
To elaborate, from the Python communicate() documentation:

> 1. send data to stdin (if input is not None);
> 2. closes stdin;
> 3. read data from stdout and stderr, until EOF is reached;
> 4. wait for process to terminate.

https://docs.python.org/3.12/library/asyncio-subprocess.html#asyncio.subprocess.Process.communicate

That means the second step causes the race condition, when piping the two processes together as in that test case.

ronf:
Interesting - it looks like closing stdin when input=None is new in 3.12, or at least it's new to do that before reading stdout/stderr. In 3.11, feeding of input to stdin happens in parallel with reading output from stdout and stderr. For cases where "input" is big enough to not all fit before the pipe blocks, it seems like the new behavior could lead to deadlocks. I'm really surprised by this change! I wonder what prompted it.

That said, thanks for this PR -- I'm happy to merge it regardless of whether asyncio keeps this change or not. It fits better actually with the explicit call to stdin.write() and the fact that it only really needs to read from stdout in this test.

AndrewAmmerlaan:
This test still fails for me in v2.14.2 with python 3.12.3.

```
================================================== FAILURES ==================================================
_______________________________ _TestAsyncFileRedirection.test_stdout_aiofile ________________________________

self = <tests.test_process._TestAsyncFileRedirection testMethod=test_stdout_aiofile>

@asynctest
async def test_stdout_aiofile(self):
"""Test with stdout redirected to an aiofile"""

data = str(id(self))

file = await aiofiles.open('stdout', 'w')

async with self.connect() as conn:
result = await conn.run('echo', input=data, stdout=file)

with open('stdout', 'r') as file:
stdout_data = file.read()

>       self.assertEqual(stdout_data, data)
E       AssertionError: '' != '140679733926784'
E       + 140679733926784

conn       = <asyncssh.connection.SSHClientConnection object at 0x7ff28cfa8560>
data       = '140679733926784'
file       = <_io.TextIOWrapper name='stdout' mode='r' encoding='UTF-8'>
result     = SSHCompletedProcess(env=mappingproxy({}), command='echo', subsystem=None, exit_status=None, exit_signal=None, returncode=None, stdout='', stderr='140679733926784')
self       = <tests.test_process._TestAsyncFileRedirection testMethod=test_stdout_aiofile>
stdout_data = ''

/tmp/portage/dev-python/asyncssh-2.14.2/work/asyncssh-2.14.2/tests/test_process.py:1180: AssertionError
--------------------------------------------- Captured log call ---------------------------------------------
```

PR #580, opened 2023-07-16
Make log show what user and host the error happened.

pete312:
Because we are running ssh to many hosts, the existing error is missing host information the permission error happened on.

It would be beneficial for that to exist in the exception and on one line of the log.  

The exception that is caught:  
`Permission denied for pete@builder1`

And log:
```
2023-07-16 07:48:04,989 asyncssh logging log line 92 INFO : [conn=0] Auth failed for pete@builder1
2023-07-16 07:48:04,990 asyncssh logging log line 92 INFO : [conn=0] Connection failure: Permission denied for pete@builder1
```

Kind Regards
Pete

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -3286,9 +3286,9 @@ class SSHClientConnection(SSHConnection):
             if self._auth:
                 return
 
-        self.logger.info('Auth failed for user %s', self._username)
+        self.logger.info('Auth failed for %s@%s' % (self._username, self._host))
 
-        self._force_close(PermissionDenied('Permission denied'))
+        self._force_close(PermissionDenied('Permission denied for %s@%s' % (self._username, self._host)))
 
     def gss_kex_auth_requested(self) -> bool:
         """Return whether to allow GSS key exchange authentication or not"""


ronf:
I'm not sure this is a good approach. There are many different ways in which a connection could fail, and you'd have to make a change like this in all of associated log calls.

The intended way to do this correlation today is via the `[conn=0]` information that's already present in the log messages. You can go backward in the log looking for earlier information associated with that connection, or with a specific channel in the case where the error happens at the channel level. For instance, here's an example log:

```
DEBUG:asyncssh:Reading config from "/Users/ronf/.ssh/config"
INFO:asyncssh:Opening SSH connection to localhost, port 22
INFO:asyncssh:[conn=0] Connected to SSH server at localhost, port 22
INFO:asyncssh:[conn=0]   Local address: ::1, port 51052
INFO:asyncssh:[conn=0]   Peer address: ::1, port 22
DEBUG:asyncssh:[conn=0] Sending version SSH-2.0-AsyncSSH_2.13.2
DEBUG:asyncssh:[conn=0] Received version SSH-2.0-OpenSSH_9.0
DEBUG:asyncssh:[conn=0] Requesting key exchange
DEBUG:asyncssh:[conn=0] Received key exchange request
DEBUG:asyncssh:[conn=0] Beginning key exchange
DEBUG:asyncssh:[conn=0] Completed key exchange
INFO:asyncssh:[conn=0] Beginning auth for user ronf
DEBUG:asyncssh:[conn=0] Trying password auth
INFO:asyncssh:[conn=0] Auth failed for user ronf
INFO:asyncssh:[conn=0] Connection failure: Permission denied
INFO:asyncssh:[conn=0] Aborting connection
```

Even if you have multiple connections running in parallel, each of them should be annotated with a connection number, an you can then look through the earlier output for this connection number to get other information like the host, address, port number, and other details such as what auth was attempted.

Here's an example where a channel is included:

```
DEBUG:asyncssh:Reading config from "/Users/ronf/.ssh/config"
INFO:asyncssh:Opening SSH connection to localhost, port 22
INFO:asyncssh:[conn=0] Connected to SSH server at localhost, port 22
INFO:asyncssh:[conn=0]   Local address: ::1, port 51220
INFO:asyncssh:[conn=0]   Peer address: ::1, port 22
DEBUG:asyncssh:[conn=0] Sending version SSH-2.0-AsyncSSH_2.13.2
DEBUG:asyncssh:[conn=0] Received version SSH-2.0-OpenSSH_9.0
DEBUG:asyncssh:[conn=0] Requesting key exchange
DEBUG:asyncssh:[conn=0] Received key exchange request
DEBUG:asyncssh:[conn=0] Beginning key exchange
DEBUG:asyncssh:[conn=0] Completed key exchange
INFO:asyncssh:[conn=0] Beginning auth for user ronf
DEBUG:asyncssh:[conn=0] Trying public key auth with ssh-ed25519-cert-v01@openssh.com key
DEBUG:asyncssh:[conn=0] Signing request with ssh-ed25519-cert-v01@openssh.com key
INFO:asyncssh:[conn=0] Auth for user ronf succeeded
INFO:asyncssh:[conn=0, chan=0] Requesting new SSH session
DEBUG:asyncssh:[conn=0] Received unknown global request: hostkeys-00@openssh.com
DEBUG:asyncssh:[conn=0, chan=0]   Terminal type: ansi
DEBUG:asyncssh:[conn=0, chan=0]   Agent forwarding enabled
INFO:asyncssh:[conn=0, chan=0]   Command: tty
INFO:asyncssh:[conn=0, chan=0] Received exit status 0
INFO:asyncssh:[conn=0] Closing connection
INFO:asyncssh:[conn=0] Sending disconnect: Disconnected by application (11)
INFO:asyncssh:[conn=0] Connection closed
```

Note that some of the lines have`[conn=0, chan=0]` on them, telling you which channel generated the log if you are opening multiple SSH sessions at once over a single connection.

If you really want the host information to appear in a specific log message without having to look backward like this, it's probably best to do that within your application. You can catch the exception in your application and print out whatever information you want in the exception handler.

pete312:
Thanks you for your speedy reply, and your dedication to this project. 

The rationale for the change (which I feel is visibility improvement) has two viewpoints. 

One is from the user perspective.  If you are looking at logs from a support user, having the context to a failure on a different line than the cause (the host in this case), is not what support groups like to deal with.   They like to search for one liners like what is suggested.

The other reasoning behind the change is the coding aspect. The exception (which is fully able to quantify the problem) is left to the client to add host and user.  If the client has not implemented the try catch with context (meaning its not able to add the host information to the call to connect) then the end user is left without any idea of how to address the issue.  Which if fine and it bubbles up to the developer to implement a try.  However the information is already available at the original exception. So force the try catch on the client code to make this info available does not make sense.

One experience note over the years, is that Permission Denied error,  is a user related problem rather than some other random system issue. So this error is not like most errors that need more investigation (support need to investigate logs). That is one reason why this is a special error with this quantifier, rather than making every log failure look like this..


One thing I could see to regard this is that if clients already code on the string of the exception. that string would change and break existing clients.  If you think this is reason that this is not a good approach I fully agree and withdraw the PR.

Kind Regard 
Pete

ronf:
For me, the main question is how one would go about debugging a problem like this. While having the remote user and remote host the connection is being made to is useful, the more important information is actually the set of auth methods attempted, and related info like which public keys auth was attempted with. To see that, you need the full debug log, and that already has the user and host information today. With only the remote user and remote host information, I don't think you're going to get very far.

There is a precedent in AsyncSSH for including extra information in the exception reason string. Specifically, this was done for KeyExchangeFailed, where the problem was that the set of key exchange algorithms requested by AsyncSSH didn't overlap at all with the set of algorithms offered by the server. So, it made sense to return the specifics of these two lists in the exception. Even without debug logging enabled, that exception provided enough information to fix the problem in many cases, as you could see things like an old server only offering something like SHA-1 algorithms that are not enabled by default any more. However, I don't think that's the case here for PermissionDenied. The remote user & remote host isn't actually enough.

Can you say a bit more about your use case? Why wouldn't the user already know the remote username and remote host that is being connected to, and what would you expect the user to do with that information if it was included when this error was raised?

pete312:
Absolutely. 

A good example is a parallel runner utility with some bells and whistles like host globbing to target sets of hosts. Its is single shot one liner parallel runner.  I based it based that tool off this example a long time ago

```
import asyncio, asyncssh

async def run_client(host, command: str) -> asyncssh.SSHCompletedProcess:
    async with asyncssh.connect(host) as conn:
        return await conn.run(command)

async def run_multiple_clients() -> None:
    hosts = ['host1', 'host2', 'host3', 'host4']

    tasks = (run_client(host, 'ls abc') for host in hosts)
    results = await asyncio.gather(*tasks, return_exceptions=True)

    for i, result in enumerate(results, 1):
        if isinstance(result, Exception):
            print('Task %d failed: %s' % (i, str(result)))
        elif result.exit_status != 0:
            print('Task %d exited with status %s:' % (i, result.exit_status))
            print(result.stderr, end='')
        else:
            print('Task %d succeeded:' % i)
            print(result.stdout, end='')

        print(75*'-')

asyncio.get_event_loop().run_until_complete(run_multiple_clients())  
```

We use host keys in general, although you can use password or public key auth methods too with the tool. As it stands, exceptions show too little information to deal with the problem when dealing with a group of failures.  When you restarting services etc.. 

```
Task 1 failed: Host key is not trusted
---------------------------------------------------------------------------
Task 2 failed: Host key is not trusted
---------------------------------------------------------------------------
Task 3 failed: Permission denied 
---------------------------------------------------------------------------
Task 4 failed: Host key is not trusted
---------------------------------------------------------------------------
```

The PR ( actually host key "host" would need to be added for this too) would improve this error message to add the minimal information without needing a logger.

```
Task 1 failed: Host key is not trusted for host1
---------------------------------------------------------------------------
Task 2 failed: Host key is not trusted for host2
---------------------------------------------------------------------------
Task 3 failed: Permission denied for user1@host3
---------------------------------------------------------------------------
Task 4 failed: Host key is not trusted for host4
---------------------------------------------------------------------------
```

Usually hostname this is enough information for this error.  The operator would then use ssh to troubleshoot connection issue rather than this script.  That seems like the most logical workflow for this.  Without this, the client code must wrap the call with a handler, to look at the error and then add onto it, a hostname.  That seems to be unnecessary boiler plate code to add to every project.

I really think this is an improvement based on the workflow I outlined.  Other apps would have a proper logger and the user would be looking at logs where this information is available in detail to your point.

Kind Regards 
pete

ronf:
Thanks. The example you described makes sense to me for the "Host key is not trusted" case, and I can see how that would also potentially apply to the "Permission denied" case.

That said, I think I would leave out the change to the debug log output. If that logging is enabled, the information would already be there, as we discussed.

The next challenge is then to determine which exceptions should have this information added and the best place to add it. What you propose in `try_next_auth` makes sense for the "Permission denied" case. However, if you want to also cover the "Host key is not trusted", it becomes a bit more complicated. Right now, I'm leaning toward using `validate_server_host_key` for this, as that covers multiple key exchange methods and validation of both host & CA keys. However, this wouldn't cover things like hash mismatch errors or various other protocol errors during key exchange, and I don't think hostname information is even available in the key exchange code right now, so we may have to rely on debug logging in this case.

Putting the change in `validate_server_host_key` also wouldn't cover the case where a compatible host key isn't found, though a similar change could be added to those places (in `_choose_alg` and `_process_kexinit` in SSHConnection, where KeyExchangeFailed is raised).

There's also another error case involving a "Trivial auth disabled" message in `_process_userauth_success` which I could see a potential argument for adding host information to.

What do you think about the different error cases here and the value of adding host information to them?

pete312:
Yes, I see the potential issue where there are many places that would potentially change. 

Example:

```
    def _validate_host_key(self, host, key_data):
           ...
           ...
            if key in self._revoked_host_keys:
                raise ValueError(f'Host key is revoked for {host}')

            if self._trusted_host_keys is not None and \
               key not in self._trusted_host_keys:
                raise ValueError(f'Host key is not trusted for {host}')
```

The `host` is available to the function here too, but I am not an expert in the protocol to say where these changes are appropriate.  But it seems sensible that where `host` is a context, the error should include the context?

ronf:
Ok - I've got a first cut at this.

Modifying `validate_server_host_key()` actually covers the `_validate_host_key` case as well as others such as validating the host with OpenSSH or X.509 certificates. In addition, this covers the `PermissionDenied` case and improves some wording on a couple of other places related to X.509 certificates I ran across while testing.

I've left out the `_choose_alg()` use case for now as that's used on both client and server and I don't think we want to change any error messages AsyncSSH sends back to remote clients when acting as a server. This could be made conditional, but it's a bit messy, especially given that `self._host` only exists in SSHClientConnection, and this method is in the parent SSHConnection class. There are some similar challenges with `_process_userauth_success()`.

If you'd like to give this change a try, here's a patch you can apply:

```patch
index 0a8d35c..a94318e 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -3240,7 +3240,12 @@ class SSHClientConnection(SSHConnection):
                 self._host_key_alias or self._host,
                 self._peer_addr, self._port, key_data)
         except ValueError as exc:
-            raise HostKeyNotVerifiable(str(exc)) from None
+            host = self._host
+
+            if self._host_key_alias:
+                host += f' with alias {self._host_key_alias}'
+
+            raise HostKeyNotVerifiable(f'{exc} for host {host}') from None

         self._server_host_key = host_key
         return host_key
@@ -3288,7 +3293,9 @@ class SSHClientConnection(SSHConnection):

         self.logger.info('Auth failed for user %s', self._username)

-        self._force_close(PermissionDenied('Permission denied'))
+        self._force_close(PermissionDenied('Permission denied for user '
+                                           f'{self._username} on host '
+                                           f'{self._host}'))

     def gss_kex_auth_requested(self) -> bool:
         """Return whether to allow GSS key exchange authentication or not"""
@@ -6548,7 +6555,8 @@ class SSHConnectionOptions(Options):
         if x509_trusted_cert_paths:
             for path in x509_trusted_cert_paths:
                 if not Path(path).is_dir():
-                    raise ValueError('Path not a directory: ' + str(path))
+                    raise ValueError('X.509 trusted certificate path not '
+                                     f'a directory: {path}')

         self.x509_trusted_certs = x509_trusted_certs
         self.x509_trusted_cert_paths = x509_trusted_cert_paths
diff --git a/asyncssh/crypto/x509.py b/asyncssh/crypto/x509.py
index c041ef2..e89fe4f 100644
--- a/asyncssh/crypto/x509.py
+++ b/asyncssh/crypto/x509.py
@@ -308,7 +308,7 @@ class X509Certificate:
                                                None)
             x509_ctx.verify_certificate()
         except crypto.X509StoreContextError as exc:
-            raise ValueError(str(exc)) from None
+            raise ValueError(f'X.509 chain validation error: {exc}') from None


 def generate_x509_certificate(signing_key: PyCAKey, key: PyCAKey,
```

ronf:
I've gone ahead and checked this into the "develop" branch as commit c3ad5f1.

pete312:
Thank you!

PR #584, opened 2023-07-30
Enable --use-pep517 for tox-invoked pip

### LINKED ISSUE #583: Workflow problem suddenly appears on macOS
@hexchain Sorry to bother you again, but I suddenly started to see workflow errors on macOS tonight which seem unrelated to the change I made. I suspect something referenced in the workflow was perhaps updated. See https://github.com/ronf/asyncssh/actions/runs/5676027821 for an example of the failures. They seem related to running "setup.py clean", getting an error of "Directory not empty". For example:

```
      OSError: [Errno 66] Directory not empty: '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-clmiyga6/python-pkcs11_3c253ba59f2343ada4601d24857ff6bd/.eggs/Cython-3.0.0-py3.7-macosx-11.7-x86_64.egg/Cython-3.0.0.dist-info' -> '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-clmiyga6/python-pkcs11_3c253ba59f2343ada4601d24857ff6bd/.eggs/Cython-3.0.0-py3.7-macosx-11.7-x86_64.egg/EGG-INFO'
```

This seems to be occurring on macOS on all Python versions (3.7 to 3.11), but not at all on Linux or Windows. It may be specific to python-pkcs11, but I'm not sure.

Any suggestions on how to get to the bottom of this?

hexchain:
Seems this is now required to make Cython (required by python-pkcs11) happy.

Closes #583.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -111,7 +111,7 @@ jobs:
           V = sys.version_info
           p = platform.system().lower()
           subprocess.run(
-              ['tox', '-e', f'py{V.major}{V.minor}-{p}', '--', '-ra'],
+              ['tox', 'run', '-e', f'py{V.major}{V.minor}-{p}', '--', '-ra'],
               check=True)
 
       - name: Upload coverage data
--- a/tox.ini
+++ b/tox.ini
@@ -26,6 +26,7 @@ platform =
     windows: win32
 usedevelop = True
 setenv =
+    PIP_USE_PEP517 = 1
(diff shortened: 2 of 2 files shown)


ronf:
Thanks so much for this! It's odd that this only affected macOS, and that I haven't been able to trigger anything like this from my local testing, which is all on macOS and should be running the latest versions of all the build tools.

PR #600, opened 2023-10-17
Use Self TypeVar for SSHConnection __aenter__

PJB3005:
This makes it so that when using a subclass as a context manager like so:

```py
async with await asyncssh.connect(...) as con:
```

the `con` variable has the type `SSHClientConnection` instead of simply `SSHConnection`.

This is equivalent to the [`typing.Self` type introduced in Python 3.11](https://docs.python.org/3.11/library/typing.html#typing.Self), but backwards compatible.

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -179,7 +179,7 @@ _ServerFactory = Callable[[], SSHServer]
 _ProtocolFactory = Union[_ClientFactory, _ServerFactory]
 
 _Conn = TypeVar('_Conn', 'SSHClientConnection', 'SSHServerConnection')
-_ConnSelf = TypeVar("_ConnSelf", bound="SSHConnection")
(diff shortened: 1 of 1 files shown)


asyncssh/connection.py
```diff
@@ -179,6 +179,7 @@
 _ProtocolFactory = Union[_ClientFactory, _ServerFactory]
 
 _Conn = TypeVar('_Conn', 'SSHClientConnection', 'SSHServerConnection')
+_ConnSelf = TypeVar("_ConnSelf", bound="SSHConnection")
```
ronf, line 182: Apologies for the slow response. Somehow I missed the notification on this.

Would you might changing this to use single quotes? That's the convention in AsyncSSH.

ronf:
Thanks very much!

PR #636, opened 2024-02-26
Treat ProxyJump as a list not a single value

### LINKED ISSUE #635: ProxyJump is not treated as a list
Hi,

In https://github.com/ronf/asyncssh/blob/develop/asyncssh/connection.py#L6996, ProxyJump is treated as if it can only contain a single hostname, but per the spec it is actually a space separated list of hosts to tunnel through. See e.g. https://man.openbsd.org/ssh_config#ProxyJump. 

A test case like the below should be sufficient to validate a solution, though setting up additional servers for the test may be prudent.
```
    @asynctest
    async def test_proxy_jump(self):
        """Test connecting a tunnneled SSH connection using ProxyJump"""

        write_file('.ssh/config', 'Host target\n'
                   '  Hostname localhost\n'
                   f'  Port {self._server_port}\n'
                   f'  ProxyJump localhost:{self._server_port} localhost:{self._server_port}\n'
                   'IdentityFile ckey\n', 'w')

        try:
            async with self.connect(host='target', username='ckey'):
                pass
        finally:
            os.remove('.ssh/config')
```
I will try to submit a draft PR with a potential fix shortly.

aiven-amartin:
The manpages and spec for OpenSSH treat ProxyJump as a list of 0 or more proxies to jump through, however the project here currently assumes exactly one value. This commit updates the config and tunnel logic to connect via multiple hosts, while still allowing for str values as users currently can manually specify a string value in a variety of cases.

fixes #635

### DIFF
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -554,7 +554,7 @@ class SSHClientConfig(SSHConfig):
         ('PreferredAuthentications',        SSHConfig._set_string),
         ('Port',                            SSHConfig._set_int),
         ('ProxyCommand',                    SSHConfig._set_string_list),
-        ('ProxyJump',                       SSHConfig._set_string),
+        ('ProxyJump',                       SSHConfig._set_string_list),
         ('PubkeyAuthentication',            SSHConfig._set_bool),
         ('RekeyLimit',                      SSHConfig._set_rekey_limits),
         ('RemoteCommand',                   SSHConfig._set_string),
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -32,7 +32,8 @@ import sys
 import tempfile
 import time
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
+from collections.abc import Iterable
 from functools import partial
 from pathlib import Path
 from types import TracebackType
@@ -233,8 +234,9 @@ _RequestPTY = Union[bool, str]
 _TCPServerHandlerFactory = Callable[[str, int], SSHSocketSessionFactory]
 _UNIXServerHandlerFactory = Callable[[], SSHSocketSessionFactory]
 
-_TunnelConnector = Union[None, str, _TunnelConnectorProtocol]
-_TunnelListener = Union[None, str, _TunnelListenerProtocol]
+_TunnelConfig = namedtuple("TunnelConfig", ["host", "port", "username"])
+_TunnelConnector = Union[None, str, list[str], _TunnelConnectorProtocol]
+_TunnelListener = Union[None, str, list[str], _TunnelListenerProtocol]
 
 _VersionArg = DefTuple[BytesOrStr]
 
@@ -347,27 +349,39 @@ async def _open_proxy(
     return cast(_Conn, cast(_ProxyCommandTunnel, tunnel).get_conn())
 
 
-async def _open_tunnel(tunnel: object, passphrase: Optional[BytesOrStr]) -> \
-        Optional['SSHClientConnection']:
-    """Parse and open connection to tunnel over"""
-
-    username: DefTuple[str]
-    port: DefTuple[int]
-
-    if isinstance(tunnel, str):
-        if '@' in tunnel:
-            username, host = tunnel.rsplit('@', 1)
+def _iterate_tunnels(tunnels: list[str]) -> Iterable[_TunnelConfig]:
+    """Iterate over a list of tunnels"""
+    for tunnel in tunnels:
+        if "@" in tunnel:
+            username, host = tunnel.rsplit("@", 1)
         else:
             username, host = (), tunnel
 
-        if ':' in host:
-            host, port_str = host.rsplit(':', 1)
+        if ":" in host:
+            host, port_str = host.rsplit(":", 1)
             port = int(port_str)
         else:
             port = ()
+        yield _TunnelConfig(host=host, port=port, username=username)
 
-        return await connect(host, port, username=username,
-                             passphrase=passphrase)
+
+async def _open_tunnel(
+    tunnel: object, passphrase: Optional[BytesOrStr]
+) -> Optional["SSHClientConnection"]:
+    """Parse and open connection to tunnel over"""
+    if isinstance(tunnel, str):
+        tunnel = [tunnel]
+    if isinstance(tunnel, list):
+        connection = None
+        for tunnel_config in _iterate_tunnels(tunnel):
+            connection = await connect(
+                host=tunnel_config.host,
+                port=tunnel_config.port,
+                username=tunnel_config.username,
+                passphrase=passphrase,
+                tunnel=connection,
+            )
+        return connection
     else:
         return None
 
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -362,6 +362,26 @@ class _TestTCPForwarding(_CheckForwarding):
             os.remove('.ssh/config')
 
     @asynctest
+    async def test_proxy_jump_multiple_jumps(self):
+        """Test connecting using ProxyJump with multiple hosts"""
+
+        write_file(
+            ".ssh/config",
+            "Host target\n"
+            "  Hostname localhost\n"
+            f"  Port {self._server_port}\n"
+            f"  ProxyJump localhost:{self._server_port} localhost:{self._server_port}\n"
+            "IdentityFile ckey\n",
+            "w",
+        )
+
+        try:
+            async with self.connect(host="target", username="ckey"):
+                pass
+        finally:
+            os.remove(".ssh/config")
+
+    @asynctest
     async def test_proxy_jump_encrypted_key(self):
         """Test ProxyJump with encrypted client key"""
 


aiven-amartin:
Note I am not sure this fix is complete, certain use cases are still failing in my environment, I am currently debugging whether those are because of incompleteness here, or incorrect calling in my code. (The failures are creating cases where connect is called with `None` as the host, so I feel like it is probably in my code, I just haven't figured out how/where yet.)

ronf:
Looking at the OpenSSH documentation and doing some testing here, it appears that ProxyJump is actually a string value, not a string list. If you want to go through multiple jump hosts, you need to specify a comma-separated list as the string. So, the code here should probably keep treating the config value as a string but do a `','.split()` on this value to turn it into a list. The case with no comma will automatically fall out here, creating a list of length 1.

I'm also thinking the iterator may not be necessary here. It may actually be less of a change to just add a loop on the split version of the string directly in place.

Also, a minor nit: AsyncSSH generally uses single quote around all string constants.

I appreciate your draft PR, but I think I'm going to take a crack at a simpler version and see if I can get that to work. The basic concept will be similar to what you're proposing here, and I'm happy to give you credit for that (as well as for noticing the issue in the first place).

 I'll let you know how it goes.

ronf:
One other complication here is that the connection object currently keeps the connection used for tunneling, so it can close both connections when the connection to final destination is closed. To support a chain here, each step in the chain will need to remember its parent tunnel, which means the `set_tunnel` call in the current code may need to be done a bit differently.

ronf:
Here's my first cut at this:

```patch
index c0db777..04b46ec 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -355,19 +355,26 @@ async def _open_tunnel(tunnel: object, passphrase: Optional[BytesOrStr]) -> \
     port: DefTuple[int]

     if isinstance(tunnel, str):
-        if '@' in tunnel:
-            username, host = tunnel.rsplit('@', 1)
-        else:
-            username, host = (), tunnel
+        conn = None

-        if ':' in host:
-            host, port_str = host.rsplit(':', 1)
-            port = int(port_str)
-        else:
-            port = ()
+        for tunnel in tunnel.split(','):
+            if '@' in tunnel:
+                username, host = tunnel.rsplit('@', 1)
+            else:
+                username, host = (), tunnel
+
+            if ':' in host:
+                host, port_str = host.rsplit(':', 1)
+                port = int(port_str)
+            else:
+                port = ()

-        return await connect(host, port, username=username,
-                             passphrase=passphrase)
+            last_conn = conn
+            conn = await connect(host, port, username=username,
+                                 passphrase=passphrase)
+            conn.set_tunnel(last_conn)
+
+        return conn
     else:
         return None

@@ -439,10 +446,6 @@ async def _connect(options: 'SSHConnectionOptions',
     try:
         await options.waiter
         free_conn = False
-
-        if new_tunnel:
-            conn.set_tunnel(new_tunnel)
-
         return conn
     finally:
         if free_conn:
```

In addition to adding support for a comma-separated list, I also cleaned up an unnecessary extra call to set_tunnel in the `_connect()` method.

aiven-amartin:
@ronf Thanks for the simpler patch, I will try it out. That said, a space delimited list is supported by at least OpenSSH 9, so would be nice to support that too, but I can change my deployments to use commas instead.

ronf:
It looks like you already edited your reply above, but just to confirm I did test out spaces instead of commas, but that didn't work for me, at least not on OpenSSH 9.6. Putting just a space seemed to cause it to ignore everything after the space so only the first jump host listed was actually used.

I also tried using a comma followed by a space, and that got me an error about a bad configuration option. The only combination I could make work was a single argument with commas between jump hosts without any spaces.

aiven-amartin:
Yeah I reviewed it and we are using commas, I was hallucinating or something when I left that earlier comment.

ronf:
No worries - thanks for confirming!

aiven-amartin:
So this seems to parse things correctly, however the host aliases are not resolved (don't think this worked before this patch either). I.e. if my config is like the below, and I do `ssh node` there will be a name resolution error for `jump1`. Conversely if there is no Proxy configuration, the name is resolved (e.g. in this case `ssh jump1` would work). I'm trying to parse through how the ssh config is parsed to look this up, but I guess a workaround is to pass in the full ip:port value in the ProxyJump section.
```
Host jump1
    Hostname 10.10.0.0
    Port 22

Host jump2
    Hostname 10.10.0.1
    Port 22

Host node
    Hostname 10.10.0.2
    Port 22
    ProxyJump jump1,jump2
```

aiven-amartin:
I guess something like this?
```
index ffe5703..f958173 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -343,27 +343,34 @@ async def _open_proxy(
     return cast(_Conn, cast(_ProxyCommandTunnel, tunnel).get_conn())


-async def _open_tunnel(tunnel: object, passphrase: Optional[BytesOrStr]) -> \
+async def _open_tunnel(tunnels: object, passphrase: Optional[BytesOrStr], tunnel_options: 'SSHConnectionOptions | None') -> \
         Optional['SSHClientConnection']:
     """Parse and open connection to tunnel over"""

     username: DefTuple[str]
     port: DefTuple[int]

-    if isinstance(tunnel, str):
-        if '@' in tunnel:
-            username, host = tunnel.rsplit('@', 1)
-        else:
-            username, host = (), tunnel
+    if isinstance(tunnels, str):
+        conn: Optional[SSHClientConnection] = None

-        if ':' in host:
-            host, port_str = host.rsplit(':', 1)
-            port = int(port_str)
-        else:
-            port = ()
+        for tunnel in tunnels.split(','):
+            if '@' in tunnel:
+                username, host = tunnel.rsplit('@', 1)
+            else:
+                username, host = (), tunnel
+
+            if ':' in host:
+                host, port_str = host.rsplit(':', 1)
+                port = int(port_str)
+            else:
+                port = ()
+
+            last_conn = conn
+            conn = await connect(host, port, username=username,
+                                 passphrase=passphrase, tunnel=conn, options=tunnel_options)
+            conn.set_tunnel(last_conn)

-        return await connect(host, port, username=username,
-                             passphrase=passphrase)
+        return conn
     else:
         return None

@@ -371,7 +378,8 @@ async def _open_tunnel(tunnel: object, passphrase: Optional[BytesOrStr]) -> \
 async def _connect(options: 'SSHConnectionOptions',
                    loop: asyncio.AbstractEventLoop, flags: int,
                    sock: Optional[socket.socket],
-                   conn_factory: Callable[[], _Conn], msg: str) -> _Conn:
+                   conn_factory: Callable[[], _Conn], msg: str,
+                   tunnel_options: 'SSHConnectionOptions | None' = None) -> _Conn:
     """Make outbound TCP or SSH tunneled connection"""

     host = options.host
@@ -384,7 +392,7 @@ async def _connect(options: 'SSHConnectionOptions',

     options.waiter = loop.create_future()

-    new_tunnel = await _open_tunnel(tunnel, options.passphrase)
+    new_tunnel = await _open_tunnel(tunnel, options.passphrase, tunnel_options)
     tunnel: _TunnelConnectorProtocol

     try:
@@ -435,10 +443,6 @@ async def _connect(options: 'SSHConnectionOptions',
     try:
         await options.waiter
         free_conn = False
-
-        if new_tunnel:
-            conn.set_tunnel(new_tunnel)
-
         return conn
     finally:
         if free_conn:
@@ -1131,7 +1135,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
                          String(self._client_kexinit),
                          String(self._server_kexinit)))

-    def set_tunnel(self, tunnel: _TunnelProtocol) -> None:
+    def set_tunnel(self, tunnel: Optional[_TunnelProtocol]) -> None:
         """Set tunnel used to open this connection"""

         self._tunnel = tunnel
@@ -8266,9 +8270,13 @@ async def connect(host = '', port: DefTuple[int] = (), *,
         host=host, port=port, tunnel=tunnel, family=family,
         local_addr=local_addr, **kwargs))

+    if new_options.tunnel:
+        tunnel_options = cast(SSHClientConnectionOptions, await _run_in_executor(
+            loop, SSHClientConnectionOptions, options, config=config, **kwargs))
+
     return await asyncio.wait_for(
         _connect(new_options, loop, flags, sock, conn_factory,
-                 'Opening SSH connection to'),
+                 'Opening SSH connection to', tunnel_options=tunnel_options),
         timeout=new_options.connect_timeout)

```

ronf:
Aliases are already working for me here, even without this change. Note how the `_open_tunnel()` calls `connect()`. This would cause it to go and read the config all over again for each proxy jump host it connects to.

Are you using a config which is stored somewhere other than the default location of ~/.ssh/config? If so, that could explain the issue. AsyncSSH follows what OpenSSH claims to do here, and doesn't use any of the options settings (including any custom config file) set for the final target when connecting to jump hosts. See the following text from the OpenSSH ssh_config man page:

            Note also that the configuration for the destination host (either
            supplied via the command-line or the configuration file) is not
            generally applied to jump hosts.  ~/.ssh/config should be used if
            specific configuration is required for jump hosts.

aiven-amartin:
Hmm, that is interesting. I don't read that paragraph the same way as you, I just see it as saying that the host specific configruation doesn't apply, but it would still respect configuration for the host in whatever configuration files exist. I am currently passing a file with `-F` that defines the aliases (no other meaningful values), but OpenSSH locally is respecting it. I don't really want to/can't pollute my home ssh config as this code will be running on servers that will connect via various gateways. That all said, even if we don't pass down the ssh_config, I still need to pass down the host_keys.

The current workflow I am trying to convert writes an ssh_config file like the above, along with a host_keys file to a temporary directory, and then passes those to openssh with a command like `ssh -F tmpdir/ssh_config node ...`. The directory looks like this:
```
tmpdir
-> ssh_config
-> known_hosts
-> node_key-cert.pub
-> node_key.pub
-> jump1_key-cert.pub
-> jump1_key.pub
-> jump2_key-cert.pub
-> jump2_key.pub
```
ssh_config looks like:
```
Host jump1
    Hostname 10.10.0.0
    Port 22
    IdentityFile tmpdir/jump1_key

Host jump2
    Hostname 10.10.0.1
    Port 22
    IdentityFile tmpdir/jump2_key

Host node
    Hostname 10.10.0.2
    Port 22
    ProxyJump jump1,jump2
    IdentityFile tmpdir/node_key
    ...

Host *
    UserKnownHostsFile tmpdir/known_hosts
```
known_hosts looks like
```
10.10.0.0 ssh-ed25519 <redacted1>
10.10.0.1 ssh-ed25519 <redacted2>
10.10.0.2 ssh-ed25519 <redacted3>
```

I don't want to just shove a bunch of this all in the ~/.ssh/config file as these keys are short lived and different threads and processes and connecting to different hosts at any given time. Am I missing an argument somewhere to override the default path? I basically want the behaviour of the F flag where if I pass in a different path, all references to `~/.ssh/config` instead reference the path provided. https://man.openbsd.org/ssh#F

aiven-amartin:
So actually I think I'm expecting behaviour to be something like 
```diff
diff --git a/asyncssh/config.py b/asyncssh/config.py
index c460817..443c678 100644
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -73,6 +73,7 @@ class SSHConfig:
         self._options = self._last_options.copy()
         self._tokens: Dict[str, str] = {}

+        self.config_paths: ConfigPaths = (self._default_path,)
         self.loaded = False

     def _error(self, reason: str, *args: object) -> NoReturn:
@@ -385,7 +386,7 @@ class SSHConfig:

             for path in paths:
                 config.parse(Path(path))
-
+            config.config_paths = paths
             config.loaded = True

         return config
diff --git a/asyncssh/connection.py b/asyncssh/connection.py
index ec56725..2ce034d 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -354,7 +354,7 @@ async def _open_proxy(
     return cast(_Conn, cast(_ProxyCommandTunnel, tunnel).get_conn())


-async def _open_tunnel(tunnels: object, passphrase: Optional[BytesOrStr]) -> \
+async def _open_tunnel(tunnels: object, passphrase: Optional[BytesOrStr], config: Optional[ConfigPaths]) -> \
         Optional['SSHClientConnection']:
     """Parse and open connection to tunnel over"""

@@ -378,7 +378,7 @@ async def _open_tunnel(tunnels: object, passphrase: Optional[BytesOrStr]) -> \

             last_conn = conn
             conn = await connect(host, port, username=username,
-                                 passphrase=passphrase, tunnel=conn)
+                                 passphrase=passphrase, tunnel=conn, config=config)
             conn.set_tunnel(last_conn)

         return conn
@@ -402,7 +402,7 @@ async def _connect(options: 'SSHConnectionOptions',

     options.waiter = loop.create_future()

-    new_tunnel = await _open_tunnel(tunnel, options.passphrase)
+    new_tunnel = await _open_tunnel(tunnel, options.passphrase, config=options.config.config_paths)
     tunnel: _TunnelConnectorProtocol

     try:
@@ -478,7 +478,7 @@ async def _listen(options: 'SSHConnectionOptions',
     tunnel = options.tunnel
     family = options.family

-    new_tunnel = await _open_tunnel(tunnel, options.passphrase)
+    new_tunnel = await _open_tunnel(tunnel, options.passphrase, config=options.config.config_paths)
     tunnel: _TunnelListenerProtocol

     if sock:

```

aiven-amartin:
Just finished a round of testing with your patch + the patch I just posted, and everything seems to work as expected.  I hope you will be amenable to this change. If so I am happy to add some tests and update this PR to be those two patches plus tests and documentation clarification.

ronf:
The `config` argument to functions like `asyncssh.connect()` will override the default config path, but currently this works only for the final connection, and not intermediate ProxyJump connections. There's a note in the `asyncssh.connect()` documentation which currently suggests calling `connect()` explicit for a jump host and then passing that connection as the `tunnel` argument to the next `connect()` when you want more fine-grained control over the options for each of those connections. However, I can see where that approach would be difficult to apply when the jump hosts are actually specified in a non-default config file.

If OpenSSH honors the config file argument for both the final destination and jump hosts, I'm willing to do that in AsyncSSH to improve compatibility.

Your proposed change looks like it will work, and I like how compact it is. However, it bothers me a bit that it is storing the paths from within the SSHConfig class and then later directly accessing them from the config object in the `_connect()` and `_listen()` functions. I'm also a bit nervous about the way config can be built up using multiple options objects, each of which might have its own config path that it loads. Here, though, we want to explicitly avoid sharing other options between the jump hosts and the final destination, even if we make an exception for the private key passphrase and now possibly for the config paths.

I think the safest thing might be to just pass through the "config" argument directly from the top-level functions. It means touching the code in more places, but I think it should be possible without touching the SSHConfig class, avoiding any potential interaction from config associated with SSHConnectionOptions objects which the caller might pass in.

I'll post an updated diff as soon as I have one.

ronf:
Here's an updated patch, including some updated documentation and a simple multi-hop tunnel test case. If you'd like to suggest other test cases testing that settings applying to opening the tunnels can come from a config file (default or otherwise), that would be welcome.

```patch
index c0db777..3174b36 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -347,32 +347,42 @@ async def _open_proxy(
     return cast(_Conn, cast(_ProxyCommandTunnel, tunnel).get_conn())


-async def _open_tunnel(tunnel: object, passphrase: Optional[BytesOrStr]) -> \
+async def _open_tunnel(tunnels: object, passphrase: Optional[BytesOrStr],
+                       config: DefTuple[ConfigPaths]) -> \
         Optional['SSHClientConnection']:
     """Parse and open connection to tunnel over"""

     username: DefTuple[str]
     port: DefTuple[int]

-    if isinstance(tunnel, str):
-        if '@' in tunnel:
-            username, host = tunnel.rsplit('@', 1)
-        else:
-            username, host = (), tunnel
+    if isinstance(tunnels, str):
+        conn: Optional[SSHClientConnection] = None

-        if ':' in host:
-            host, port_str = host.rsplit(':', 1)
-            port = int(port_str)
-        else:
-            port = ()
+        for tunnel in tunnels.split(','):
+            if '@' in tunnel:
+                username, host = tunnel.rsplit('@', 1)
+            else:
+                username, host = (), tunnel
+
+            if ':' in host:
+                host, port_str = host.rsplit(':', 1)
+                port = int(port_str)
+            else:
+                port = ()

-        return await connect(host, port, username=username,
-                             passphrase=passphrase)
+            last_conn = conn
+            conn = await connect(host, port, username=username,
+                                 passphrase=passphrase, tunnel=conn,
+                                 config=config)
+            conn.set_tunnel(last_conn)
+
+        return conn
     else:
         return None


 async def _connect(options: 'SSHConnectionOptions',
+                   config: DefTuple[ConfigPaths],
                    loop: asyncio.AbstractEventLoop, flags: int,
                    sock: Optional[socket.socket],
                    conn_factory: Callable[[], _Conn], msg: str) -> _Conn:
@@ -388,7 +398,7 @@ async def _connect(options: 'SSHConnectionOptions',

     options.waiter = loop.create_future()

-    new_tunnel = await _open_tunnel(tunnel, options.passphrase)
+    new_tunnel = await _open_tunnel(tunnel, options.passphrase, config)
     tunnel: _TunnelConnectorProtocol

     try:
@@ -439,10 +449,6 @@ async def _connect(options: 'SSHConnectionOptions',
     try:
         await options.waiter
         free_conn = False
-
-        if new_tunnel:
-            conn.set_tunnel(new_tunnel)
-
         return conn
     finally:
         if free_conn:
@@ -451,6 +457,7 @@ async def _connect(options: 'SSHConnectionOptions',


 async def _listen(options: 'SSHConnectionOptions',
+                  config: DefTuple[ConfigPaths],
                   loop: asyncio.AbstractEventLoop, flags: int,
                   backlog: int, sock: Optional[socket.socket],
                   reuse_address: bool, reuse_port: bool,
@@ -468,7 +475,7 @@ async def _listen(options: 'SSHConnectionOptions',
     tunnel = options.tunnel
     family = options.family

-    new_tunnel = await _open_tunnel(tunnel, options.passphrase)
+    new_tunnel = await _open_tunnel(tunnel, options.passphrase, config)
     tunnel: _TunnelListenerProtocol

     if sock:
@@ -1140,7 +1147,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
                          String(self._client_kexinit),
                          String(self._server_kexinit)))

-    def set_tunnel(self, tunnel: _TunnelProtocol) -> None:
+    def set_tunnel(self, tunnel: Optional[_TunnelProtocol]) -> None:
         """Set tunnel used to open this connection"""

         self._tunnel = tunnel
@@ -8425,7 +8432,7 @@ async def run_client(sock: socket.socket, config: DefTuple[ConfigPaths] = (),
         loop, SSHClientConnectionOptions, options, config=config, **kwargs))

     return await asyncio.wait_for(
-        _connect(new_options, loop, 0, sock, conn_factory,
+        _connect(new_options, config, loop, 0, sock, conn_factory,
                  'Starting SSH client on'),
         timeout=new_options.connect_timeout)

@@ -8474,7 +8481,7 @@ async def run_server(sock: socket.socket, config: DefTuple[ConfigPaths] = (),
         loop, SSHServerConnectionOptions, options, config=config, **kwargs))

     return await asyncio.wait_for(
-        _connect(new_options, loop, 0, sock, conn_factory,
+        _connect(new_options, config, loop, 0, sock, conn_factory,
                  'Starting SSH server on'),
         timeout=new_options.connect_timeout)

@@ -8525,15 +8532,16 @@ async def connect(host = '', port: DefTuple[int] = (), *,
            over this connection to the requested host and port rather than
            connecting directly via TCP. A string of the form
            [user@]host[:port] may also be specified, in which case a
-           connection will first be made to that host and it will then be
-           used as a tunnel.
+           connection will be made to that host and then used as a tunnel.
+           A comma-separated list may also be specified to establish a
+           tunnel through multiple hosts.

                .. note:: When specifying tunnel as a string, any config
                          options in the call will apply only when opening
-                         the connection inside the tunnel. The tunnel
-                         itself will be opened with default configuration
-                         settings or settings in the default config file.
-                         To get more control of config settings used to
+                         a connection to the final destination host and
+                         port. However, settings to use when opening
+                         tunnels may be specified via a configuration file.
+                         To get more control of config options used to
                          open the tunnel, :func:`connect` can be called
                          explicitly, and the resulting client connection
                          can be passed as the tunnel argument.
@@ -8593,7 +8601,7 @@ async def connect(host = '', port: DefTuple[int] = (), *,
         local_addr=local_addr, **kwargs))

     return await asyncio.wait_for(
-        _connect(new_options, loop, flags, sock, conn_factory,
+        _connect(new_options, config, loop, flags, sock, conn_factory,
                  'Opening SSH connection to'),
         timeout=new_options.connect_timeout)

@@ -8631,15 +8639,16 @@ async def connect_reverse(
            over this connection to the requested host and port rather than
            connecting directly via TCP. A string of the form
            [user@]host[:port] may also be specified, in which case a
-           connection will first be made to that host and it will then be
-           used as a tunnel.
+           connection will be made to that host and then used as a tunnel.
+           A comma-separated list may also be specified to establish a
+           tunnel through multiple hosts.

                .. note:: When specifying tunnel as a string, any config
                          options in the call will apply only when opening
-                         the connection inside the tunnel. The tunnel
-                         itself will be opened with default configuration
-                         settings or settings in the default config file.
-                         To get more control of config settings used to
+                         a connection to the final destination host and
+                         port. However, settings to use when opening
+                         tunnels may be specified via a configuration file.
+                         To get more control of config options used to
                          open the tunnel, :func:`connect` can be called
                          explicitly, and the resulting client connection
                          can be passed as the tunnel argument.
@@ -8694,7 +8703,7 @@ async def connect_reverse(
         local_addr=local_addr, **kwargs))

     return await asyncio.wait_for(
-        _connect(new_options, loop, flags, sock, conn_factory,
+        _connect(new_options, config, loop, flags, sock, conn_factory,
                  'Opening reverse SSH connection to'),
         timeout=new_options.connect_timeout)

@@ -8723,20 +8732,21 @@ async def listen(host = '', port: DefTuple[int] = (), *,
            The port number to listen on. If not specified, the default
            SSH port is used.
        :param tunnel: (optional)
-           An existing SSH client connection that this new listener should
-           be forwarded over. If set, a remote TCP/IP listener will be
-           opened on this connection on the requested host and port rather
-           than listening directly via TCP. A string of the form
+           An existing SSH client connection that this new connection should
+           be tunneled over. If set, a direct TCP/IP tunnel will be opened
+           over this connection to the requested host and port rather than
+           connecting directly via TCP. A string of the form
            [user@]host[:port] may also be specified, in which case a
-           connection will first be made to that host and it will then be
-           used as a tunnel.
+           connection will be made to that host and then used as a tunnel.
+           A comma-separated list may also be specified to establish a
+           tunnel through multiple hosts.

                .. note:: When specifying tunnel as a string, any config
                          options in the call will apply only when opening
-                         the connection inside the tunnel. The tunnel
-                         itself will be opened with default configuration
-                         settings or settings in the default config file.
-                         To get more control of config settings used to
+                         a connection to the final destination host and
+                         port. However, settings to use when opening
+                         tunnels may be specified via a configuration file.
+                         To get more control of config options used to
                          open the tunnel, :func:`connect` can be called
                          explicitly, and the resulting client connection
                          can be passed as the tunnel argument.
@@ -8817,7 +8827,7 @@ async def listen(host = '', port: DefTuple[int] = (), *,
     new_options.proxy_command = None

     return await asyncio.wait_for(
-        _listen(new_options, loop, flags, backlog, sock, reuse_address,
+        _listen(new_options, config, loop, flags, backlog, sock, reuse_address,
                 reuse_port, conn_factory, 'Creating SSH listener on'),
         timeout=new_options.connect_timeout)

@@ -8857,20 +8867,21 @@ async def listen_reverse(host = '', port: DefTuple[int] = (), *,
            The port number to listen on. If not specified, the default
            SSH port is used.
        :param tunnel: (optional)
-           An existing SSH client connection that this new listener should
-           be forwarded over. If set, a remote TCP/IP listener will be
-           opened on this connection on the requested host and port rather
-           than listening directly via TCP. A string of the form
+           An existing SSH client connection that this new connection should
+           be tunneled over. If set, a direct TCP/IP tunnel will be opened
+           over this connection to the requested host and port rather than
+           connecting directly via TCP. A string of the form
            [user@]host[:port] may also be specified, in which case a
-           connection will first be made to that host and it will then be
-           used as a tunnel.
+           connection will be made to that host and then used as a tunnel.
+           A comma-separated list may also be specified to establish a
+           tunnel through multiple hosts.

                .. note:: When specifying tunnel as a string, any config
                          options in the call will apply only when opening
-                         the connection inside the tunnel. The tunnel
-                         itself will be opened with default configuration
-                         settings or settings in the default config file.
-                         To get more control of config settings used to
+                         a connection to the final destination host and
+                         port. However, settings to use when opening
+                         tunnels may be specified via a configuration file.
+                         To get more control of config options used to
                          open the tunnel, :func:`connect` can be called
                          explicitly, and the resulting client connection
                          can be passed as the tunnel argument.
@@ -8955,7 +8966,7 @@ async def listen_reverse(host = '', port: DefTuple[int] = (), *,
     new_options.proxy_command = None

     return await asyncio.wait_for(
-        _listen(new_options, loop, flags, backlog, sock,
+        _listen(new_options, config, loop, flags, backlog, sock,
                 reuse_address, reuse_port, conn_factory,
                 'Creating reverse direction SSH listener on'),
         timeout=new_options.connect_timeout)
@@ -9043,15 +9054,16 @@ async def get_server_host_key(
            over this connection to the requested host and port rather than
            connecting directly via TCP. A string of the form
            [user@]host[:port] may also be specified, in which case a
-           connection will first be made to that host and it will then be
-           used as a tunnel.
+           connection will be made to that host and then used as a tunnel.
+           A comma-separated list may also be specified to establish a
+           tunnel through multiple hosts.

                .. note:: When specifying tunnel as a string, any config
                          options in the call will apply only when opening
-                         the connection inside the tunnel. The tunnel
-                         itself will be opened with default configuration
-                         settings or settings in the default config file.
-                         To get more control of config settings used to
+                         a connection to the final destination host and
+                         port. However, settings to use when opening
+                         tunnels may be specified via a configuration file.
+                         To get more control of config options used to
                          open the tunnel, :func:`connect` can be called
                          explicitly, and the resulting client connection
                          can be passed as the tunnel argument.
@@ -9135,7 +9147,7 @@ async def get_server_host_key(
         kex_algs=kex_algs, client_version=client_version))

     conn = await asyncio.wait_for(
-        _connect(new_options, loop, flags, sock, conn_factory,
+        _connect(new_options, config, loop, flags, sock, conn_factory,
                  'Fetching server host key from'),
         timeout=new_options.connect_timeout)

@@ -9185,15 +9197,16 @@ async def get_server_auth_methods(
            over this connection to the requested host and port rather than
            connecting directly via TCP. A string of the form
            [user@]host[:port] may also be specified, in which case a
-           connection will first be made to that host and it will then be
-           used as a tunnel.
+           connection will be made to that host and then used as a tunnel.
+           A comma-separated list may also be specified to establish a
+           tunnel through multiple hosts.

                .. note:: When specifying tunnel as a string, any config
                          options in the call will apply only when opening
-                         the connection inside the tunnel. The tunnel
-                         itself will be opened with default configuration
-                         settings or settings in the default config file.
-                         To get more control of config settings used to
+                         a connection to the final destination host and
+                         port. However, settings to use when opening
+                         tunnels may be specified via a configuration file.
+                         To get more control of config options used to
                          open the tunnel, :func:`connect` can be called
                          explicitly, and the resulting client connection
                          can be passed as the tunnel argument.
@@ -9278,7 +9291,7 @@ async def get_server_auth_methods(
         client_version=client_version))

     conn = await asyncio.wait_for(
-        _connect(new_options, loop, flags, sock, conn_factory,
+        _connect(new_options, config, loop, flags, sock, conn_factory,
                  'Fetching server auth methods from'),
         timeout=new_options.connect_timeout)

diff --git a/tests/test_forward.py b/tests/test_forward.py
index f570c40..420f251 100644
--- a/tests/test_forward.py
+++ b/tests/test_forward.py
@@ -361,6 +361,23 @@ class _TestTCPForwarding(_CheckForwarding):
         finally:
             os.remove('.ssh/config')

+    @asynctest
+    async def test_proxy_jump_multiple(self):
+        """Test connecting a tunnneled SSH connection using ProxyJump"""
+
+        write_file('.ssh/config', 'Host target\n'
+                   '  Hostname localhost\n'
+                   f'  Port {self._server_port}\n'
+                   f'  ProxyJump localhost:{self._server_port},'
+                   f'localhost:{self._server_port}\n'
+                   'IdentityFile ckey\n', 'w')
+
+        try:
+            async with self.connect(host='target', username='ckey'):
+                pass
+        finally:
+            os.remove('.ssh/config')
+
     @asynctest
     async def test_proxy_jump_encrypted_key(self):
         """Test ProxyJump with encrypted client key"""
```

ronf:
Have you had a chance to try this out? If it is working for you, I'd like to commit it.

aiven-amartin:
@ronf Hi sorry I haven't had time to cut things over to your version yet, things are running smoothly with my version of the patch, will get back to rebuilding soon. The issue is your patch does not apply cleanly to the version of asyncssh run in part of our environment, so I have to tweak multiple versions of it for different versions of asyncssh (Fedora 38 and 39 ship different versions, F39 I think is new enough your patch applies cleanly, but F38 is on 2.12 vs 2.14.2 on F39).

ronf:
This change is now in the "develop" branch as commit f7b2992.

aiven-amartin:
FYI we have now deployed your version of the patch into our F39 environments, and no issues reported so far!

ronf:
Thanks for the confirmation -- this feature will be included in the next release...

PR #639, opened 2024-03-04
Add example without context manager.

### LINKED ISSUE #638: Make it clear that _ACMWrapper is awaitable.
When I first tried to use this library I was thrown by the `_ACMWrapper` return type. I didn't realise / it wasn't clear from the docs that it implemented `__await__` and simply passed through the connection object. My editor showed me the return type of `asyncssh.connect()` and I became confused as it's not consistent with [the docs](https://asyncssh.readthedocs.io/en/stable/api.html#asyncssh.connect) stating that it returns `SSHClientConnection`.

Looking at the [client examples](https://asyncssh.readthedocs.io/en/stable/#client-examples), there isn't an example of using `await asyncssh.connect()`. I also can't find that string by searching the docs.

I would propose adding 2 versions of the the [Simple Client](https://asyncssh.readthedocs.io/en/stable/#simple-client) example. One using a context manager and one using `await asyncssh.connect()`. This is a good place to do so because this is the first place the user is directed after opening the docs.

Why is this a good idea? Because as it stands, the only way to figure this out is to either:
* Read the source code and track down [this line](https://github.com/ronf/asyncssh/blob/21a3062557f334d6183af4946954b0eb6e81aba9/asyncssh/misc.py#L270)
* Find [this page](https://asyncssh.readthedocs.io/en/stable/api.html#asyncssh.connect) in the API documentation.

mjmdavis:
closes #638

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,6 +30,22 @@ asyncio framework.
 
 Check out the `examples`__ to get started!
 
+
+.. code:: python
+
+  import asyncio, asyncssh, sys
+
+  async def run_client():
+      client = asyncssh.connect('localhost')
+      result = await conn.run('echo "Hello!"', check=True)
+      print(result.stdout, end='')
+      client.close()
+
+  try:
+      asyncio.get_event_loop().run_until_complete(run_client())
+  except (OSError, asyncssh.Error) as exc:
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for the suggestion. A new example of this was added in commit a3c3fce.

PR #656, opened 2024-05-14
Have SSHProcess's __aenter__ return type be Self

MBradbury:
Set the return type of `SSHProcess.__aenter__` to be `Self`.

Without this a `SSHClientProcess` used in a context manager the variable type can decay to a `SSHProcess`.

I still need to support Python 3.10, which is why Self has come from `typing_extensions` and not `typing`.

### DIFF
--- a/asyncssh/process.py
+++ b/asyncssh/process.py
@@ -33,7 +33,7 @@ from types import TracebackType
 from typing import Any, AnyStr, Awaitable, Callable, Dict, Generic, IO
 from typing import Iterable, List, Mapping, Optional, Set, TextIO
 from typing import Tuple, Type, TypeVar, Union, cast
-from typing_extensions import Protocol
+from typing_extensions import Protocol, Self
 
 from .channel import SSHChannel, SSHClientChannel, SSHServerChannel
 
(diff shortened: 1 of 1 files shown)


ronf:
Right now, AsyncSSH only requires version 3.6 or later of typing_extensions, and it looks like Self didn't appear in typing_extensions until version 4.0.0 (and as you noted not in Python itself until 3.11).

To use "Self", AsyncSSH will need its setup.py to be modified to reflect the new typing_extensions version requirement and it will need to be tested since this is a backwards-incompatible version change in typing_extensions.

Note that a change like this was made for SSHConnection some time ago using:

```python
_ConnSelf = TypeVar('_ConnSelf', bound='SSHConnection')

    async def __aenter__(self: _ConnSelf) -> _ConnSelf:
        """Allow SSHConnection to be used as an async context manager"""

        return self
```

It would probably make sense to switch this to actually use "Self" (and possibly do so for other classes with `__aenter__`) if we move up the required version of typing_extensions. Alternately, if there are compatibility problems, the trick above could be used in `SSHProcess` without needing to update the typing_extensions version requirement.

ronf:
I did some testing this weekend and didn't run into any issues in bumping the minimum typing_extensions version to 4.0.0. So, I've gone ahead and committed 38820c7 in the "develop" branch to begin using `Self` in all the classes where `__aenter__` is defined. This includes removing the above `_ConnSelf` definition and replacing it with `Self`.

MBradbury:
Thanks, sorry I had meant to get back to this. I did also try removing the annotation, this seemed to let pyright correctly type the return of self.

ronf:
No problem. If the change I made here doesn't work for you, let me know. It should be a superset of the PR you proposed, though, just extending to more classes.

PR #671, opened 2024-07-15
Add typing default for SFTPClient.listdir

tjstum:
At runtime, it's possible to call `listdir()` (i.e. without arguments) and get a `Sequence[str]` back, due to the default argument in the implementation. However, there is no `@overload` that contains 0 arguments, so mypy doesn't think that it's possible.

This adds a simple overload for that case.

### DIFF
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -5168,7 +5168,7 @@ class SFTPClient:
         Sequence[bytes]: ... # pragma: no cover
 
     @overload
-    async def listdir(self, path: FilePath) -> \
+    async def listdir(self, path: FilePath = ...) -> \
         Sequence[str]: ... # pragma: no cover
 
     async def listdir(self, path: _SFTPPath = '.') -> Sequence[BytesOrStr]:


ronf:
Thanks for the report!

I'm actually not able to get mypy to report this error, but from looking at the code, I agree there could be an issue here.

Rather than adding a whole new overload, does it work if you make the following change?

```patch
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -5166,7 +5166,7 @@ class SFTPClient:
         Sequence[bytes]: ... # pragma: no cover

     @overload
-    async def listdir(self, path: FilePath) -> \
+    async def listdir(self, path: FilePath = ...) -> \
         Sequence[str]: ... # pragma: no cover

     async def listdir(self, path: _SFTPPath = '.') -> Sequence[BytesOrStr]:
```

I think that would be a cleaner way to represent what's actually happening here.

tjstum:
I tested this in `examples/sftp_client.py` by adding a call to `await sftp.listdir()`. When I run `mypy examples/sftp_client.py` without this change, I get this error:
```
$ mypy examples/sftp_client.py
examples/sftp_client.py:28: error: All overload variants of "listdir" of "SFTPClient" require at least one argument  [call-overload]
examples/sftp_client.py:28: note: Possible overload variants:
examples/sftp_client.py:28: note:     def listdir(self, path: bytes) -> Coroutine[Any, Any, Sequence[bytes]]
examples/sftp_client.py:28: note:     def listdir(self, path: str | PurePath) -> Coroutine[Any, Any, Sequence[str]]
Found 1 error in 1 file (checked 1 source file)
```

Adding the default argument to the second overload does also fix the issue, though (I confirmed with mypy's `reveal_type`), so I'll go with that approach!

ronf:
> I tested this in examples/sftp_client.py by adding a call to await sftp.listdir().

Yeah - that's exactly what I tried here, but I wasn't able to get it to generate an error. I even updated to the latest mypy (1.10.1) and still didn't see an error.

That said, I'm happy to take the fix.

ronf:
Sorry -- after committing this, I noticed it was done against "master". I generally prefer to keep all development in the "develop" branch until it is time to cut a new release, so I rolled that back. It's a simple enough change that I can just add it to "develop" myself, or if you like I can take a new PR from you against that branch. Let me know which you'd prefer.

tjstum:
Yup no worries. I couldn't quite understand from the CONTRIBUTING doc if you'd want this against `develop` or `master`, since it felt more like a bugfix than new development. I resubmitted this just now.

PR #673, opened 2024-07-17
Add default for SFTPClient.listdir

tjstum:
At runtime, it's possible to call `listdir()` (i.e. without arguments) and get a `Sequence[str]` back, due to the default argument in the implementation. However, there is no `@overload` that contains 0 arguments, so mypy doesn't think that it's possible.

This adds the presence of the default to the sequence-of-strings returning overload, to match the implementation

(This is a resubmit of #671 but based on `develop` instead of `master`)

### DIFF
--- a/asyncssh/sftp.py
+++ b/asyncssh/sftp.py
@@ -5168,7 +5168,7 @@ class SFTPClient:
         Sequence[bytes]: ... # pragma: no cover
 
     @overload
-    async def listdir(self, path: FilePath) -> \
+    async def listdir(self, path: FilePath = ...) -> \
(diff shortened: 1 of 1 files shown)


ronf:
Thanks very much! This is now merged.

Also, thanks for pointing out the text about the branches in CONTRIBUTING.rst. In the early days of AsyncSSH, I did accept small fixes directly in the "master" branch, but it ended up making the branch history messier when it came time to move changes from "develop" to "master". I later switched to having "develop" be the default branch in GitHub and having all changes go in there first, but I never updated CONTRIBUTING.rst to reflect that change. This is now fixed.

PR #679, opened 2024-08-18
Deduplicate results from getaddrinfo

cjwatson:
In certain configurations it's possible to get duplicate results back from `getaddrinfo`: for example, if you accidentally have more than one line in `/etc/hosts` mapping the same name to the same IP address, then Linux/glibc systems will return multiple identical entries.  This minor misconfiguration is normally harmless, but it caused `create_tcp_local_listener` to fail due to trying to bind the same address multiple times, which in turn caused the X11 tests to fail in a confusing way.  It's easy enough to deduplicate these.

Fixes https://bugs.debian.org/1052788.

### DIFF
--- a/asyncssh/listener.py
+++ b/asyncssh/listener.py
@@ -25,7 +25,7 @@ import errno
 import socket
 from types import TracebackType
 from typing import TYPE_CHECKING, AnyStr, Callable, Generic, List, Optional
-from typing import Sequence, Tuple, Type, Union
+from typing import Sequence, Set, Tuple, Type, Union
 from typing_extensions import Self
 
 from .forward import SSHForwarderCoro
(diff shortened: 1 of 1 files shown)


ronf:
Thanks for the report, and the proposed fix. I rearranged the code slightly here, since the unpacked addrinfo_entry is not needed until after we decide whether it is a duplicate or not. I also added a comment about why this is needed:

```patch
diff --git a/asyncssh/listener.py b/asyncssh/listener.py
index 7c25508..c281413 100644
--- a/asyncssh/listener.py
+++ b/asyncssh/listener.py
@@ -25,7 +25,7 @@ import errno
 import socket
 from types import TracebackType
 from typing import TYPE_CHECKING, AnyStr, Callable, Generic, List, Optional
-from typing import Sequence, Tuple, Type, Union
+from typing import Sequence, Set, Tuple, Type, Union
 from typing_extensions import Self

 from .forward import SSHForwarderCoro
@@ -285,9 +285,19 @@ async def create_tcp_local_listener(
     if not addrinfo: # pragma: no cover
         raise OSError('getaddrinfo() returned empty list')

+    seen_addrinfo: Set[Tuple] = set()
     servers: List[asyncio.AbstractServer] = []

-    for family, socktype, proto, _, sa in addrinfo:
+    for addrinfo_entry in addrinfo:
+        # Work around an issue where getaddrinfo() on some systems may
+        # return duplicate results, causing bind to fail.
+        if addrinfo_entry in seen_addrinfo: # pragma: no cover
+            continue
+
+        seen_addrinfo.add(addrinfo_entry)
+
+        family, socktype, proto, _, sa = addrinfo_entry
+
         try:
             sock = socket.socket(family, socktype, proto)
         except OSError: # pragma: no cover
```

This is checked into the "develop" branch as commit f2912e8.

PR #694, opened 2024-09-29
drop python<=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been
EOSed 27 Jun 2023.
Filter all code over `pyupgracde --py38-plus`.


kloczek:
BTW next month python 3.8 will be EOSed as well.

ronf:
I ran my own scans here with pyupgrade and ruff, but keeping support for 3.7 for now. The results are in commit 09e2fc5, and I also did a separate pass at converting to f-strings in most places in commit 25cb98f.

Thanks for the suggestion!

PR #697, opened 2024-10-07
Use bcrypt rather than crypt in simple_server example

cjwatson:
`crypt` was removed from Python 3.13.  `bcrypt` isn't ideal, but it has acceptable password hashing, is simple to use, and is already an optional dependency of asyncssh.


ronf:
Thanks for this, Colin!

As you said, bcrypt isn't ideal here but it's probably simpler than most other alternatives. This was meant to be a working example which included basic password validation, but I don't really expect someone to use this example in production code.

PR #715, opened 2024-11-17
Move project metadata to pyproject.toml

cdce8p:
The metadata changes:

```diff
 Metadata-Version: 2.1
 Name: asyncssh
 Version: 2.18.0
 Summary: AsyncSSH: Asynchronous SSHv2 client and server library
-Author: Ron Frederick
-Author-email: ronf@timeheart.net
+Author-email: Ron Frederick <ronf@timeheart.net>
-License: Eclipse Public License v2.0
+License: EPL-2.0 OR GPL-2.0-or-later
-Home-page: http://asyncssh.timeheart.net
+Project-URL: Homepage, http://asyncssh.timeheart.net
 Project-URL: Documentation, https://asyncssh.readthedocs.io
 Project-URL: Source, https://github.com/ronf/asyncssh
 Project-URL: Tracker, https://github.com/ronf/asyncssh/issues
-Platform: Any
 Classifier: Development Status :: 5 - Production/Stable
 Classifier: Environment :: Console
 Classifier: Intended Audience :: Developers
 Classifier: License :: OSI Approved
 Classifier: Operating System :: MacOS :: MacOS X
 Classifier: Operating System :: POSIX
 Classifier: Programming Language :: Python :: 3.7
 Classifier: Programming Language :: Python :: 3.8
 Classifier: Programming Language :: Python :: 3.9
 Classifier: Programming Language :: Python :: 3.10
 Classifier: Programming Language :: Python :: 3.11
 Classifier: Programming Language :: Python :: 3.12
 Classifier: Topic :: Internet
 Classifier: Topic :: Security :: Cryptography
 Classifier: Topic :: Software Development :: Libraries :: Python Modules
 Classifier: Topic :: System :: Networking
-Requires-Python: >= 3.6
+Requires-Python: >=3.6
+Description-Content-Type: text/x-rst
 License-File: LICENSE
 Requires-Dist: cryptography >=39.0
 Requires-Dist: typing-extensions >=4.0.0
 Provides-Extra: bcrypt
 Requires-Dist: bcrypt >=3.1.3 ; extra == 'bcrypt'
 Provides-Extra: fido2
 Requires-Dist: fido2 >=0.9.2 ; extra == 'fido2'
 Provides-Extra: gssapi
 Requires-Dist: gssapi >=1.2.0 ; extra == 'gssapi'
 Provides-Extra: libnacl
 Requires-Dist: libnacl >=1.4.2 ; extra == 'libnacl'
 Provides-Extra: pkcs11
 Requires-Dist: python-pkcs11 >=0.7.0 ; extra == 'pkcs11'
 Provides-Extra: pyopenssl
 Requires-Dist: pyOpenSSL >=23.0.0 ; extra == 'pyopenssl'
 Provides-Extra: pywin32
 Requires-Dist: pywin32 >=227 ; extra == 'pywin32'

 ...
```

- **License**: Use the SPDX license identifier
- **Platform**: Removed as it isn't really used for anything
- Removed the static `__author__`, `__author_email__` and `__url__` from `asyncssh/version.py`. These attributes can be accessed with `importlib.metadata`.


ronf:
Thanks for this. I see a couple of issues, though:

* I don't think relying on importlib metadata is a good idea here, as it looks like that didn't appear in the Python standard library until Python 3.8.

* The `__author__` variable is used by SFTP to fill in the vendor name for the vendor_id extension. So, that needs to be available programmatically similar to `__version__`. While there's no such requirement for `__author_email__` right now, I'm thinking it would probably be best to leave version.py unchanged and continue to pull everything from there as well just to keep these values together.

cdce8p:
I reverted the changes related to `asyncssh/version.py`. The new metadata diff:

```diff
 Metadata-Version: 2.1
 Name: asyncssh
 Version: 2.18.0
 Summary: AsyncSSH: Asynchronous SSHv2 client and server library
 Home-page: http://asyncssh.timeheart.net
 Author: Ron Frederick
 Author-email: ronf@timeheart.net
-License: Eclipse Public License v2.0
+License: EPL-2.0 OR GPL-2.0-or-later
 Project-URL: Homepage, http://asyncssh.timeheart.net
 Project-URL: Documentation, https://asyncssh.readthedocs.io
 Project-URL: Source, https://github.com/ronf/asyncssh
 Project-URL: Tracker, https://github.com/ronf/asyncssh/issues
-Platform: Any
 Classifier: Development Status :: 5 - Production/Stable
 Classifier: Environment :: Console
 Classifier: Intended Audience :: Developers
 Classifier: License :: OSI Approved
 Classifier: Operating System :: MacOS :: MacOS X
 Classifier: Operating System :: POSIX
 Classifier: Programming Language :: Python :: 3.7
 Classifier: Programming Language :: Python :: 3.8
 Classifier: Programming Language :: Python :: 3.9
 Classifier: Programming Language :: Python :: 3.10
 Classifier: Programming Language :: Python :: 3.11
 Classifier: Programming Language :: Python :: 3.12
 Classifier: Topic :: Internet
 Classifier: Topic :: Security :: Cryptography
 Classifier: Topic :: Software Development :: Libraries :: Python Modules
 Classifier: Topic :: System :: Networking
-Requires-Python: >= 3.6
+Requires-Python: >=3.6
+Description-Content-Type: text/x-rst
 License-File: LICENSE
 Requires-Dist: cryptography >=39.0
 Requires-Dist: typing-extensions >=4.0.0
 Provides-Extra: bcrypt
 Requires-Dist: bcrypt >=3.1.3 ; extra == 'bcrypt'
 Provides-Extra: fido2
 Requires-Dist: fido2 >=0.9.2 ; extra == 'fido2'
 Provides-Extra: gssapi
 Requires-Dist: gssapi >=1.2.0 ; extra == 'gssapi'
 Provides-Extra: libnacl
 Requires-Dist: libnacl >=1.4.2 ; extra == 'libnacl'
 Provides-Extra: pkcs11
 Requires-Dist: python-pkcs11 >=0.7.0 ; extra == 'pkcs11'
 Provides-Extra: pyopenssl
 Requires-Dist: pyOpenSSL >=23.0.0 ; extra == 'pyopenssl'
 Provides-Extra: pywin32
 Requires-Dist: pywin32 >=227 ; extra == 'pywin32'

 ...
```

setup.py
```diff
@@ -34,59 +34,9 @@
 
 base_dir = path.abspath(path.dirname(__file__))
 
-doclines = __doc__.split('\n', 1)
```
cdce8p, line 37: Should I remove the docsting in `setup.py`? It isn't used anymore for the project metadata.

ronf: I see the long description has been replaced with a direct reference to README.rst, which is definitely nicer, and moving the single-line description directly into pyproject.toml seems fine. So, yes, I think getting rid of the doc string in setup.py makes sense. Is there any way to avoid having the file entirely, more like your last change? That would be preferable.

You managed to dynamically access version in the original commit. Couldn't something similar work for author & email? That should also eliminate the need to read & exec version.py. As for __url__, it looks like it may not be needed any in either setup.py or pyproject.toml, but for the sake of this change I'd leave __url__ in version.py as-is.

cdce8p: > I think getting rid of the doc string in setup.py makes sense.

👍🏻 Done

> Is there any way to avoid having the file entirely, more like your last change? That would be preferable.
> 
> You managed to dynamically access version in the original commit. Couldn't something similar work for author & email? That should also eliminate the need to read & exec version.py.

Unfortunately, not that I know of. Setuptools only supports dynamic metadata for a few select fields.
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

The "best" alternative would probably be to either use `importlib.metadata` as suggested before or just duplicate the information. It doesn't really change so there wouldn't be much risk with both diverting.

> As for **url**, it looks like it may not be needed any in either setup.py or pyproject.toml, but for the sake of this change I'd leave **url** in version.py as-is.

`url` is actually used. Either for `Home-page` in `setup.py` or the `Project-URL: Homepage` in `pyproject.toml`. I can stay as is though for the time being.
```diff
-Home-page: http://asyncssh.timeheart.net
+Project-URL: Homepage, http://asyncssh.timeheart.net
```

ronf: In that case, I think your original commit with setup.py deleted is probably the best option, but with the change in version.py reverted so all the variables remain set there. As you said, there's a small amount of duplication there, but that looks a lot cleaner overall.

ronf:
This change is now in the "develop" branch, and will be included in the next release. Thanks very much for your work on this, Marc!

PR #718, opened 2024-11-21
Store references to tasks

### LINKED ISSUE #716: SSHConnection does not save a reference to tasks it creates
This function:
https://github.com/ronf/asyncssh/blob/ca15e1123ff335feea5a74926feb97a8f04986cb/asyncssh/connection.py#L1097-L1104
is used to schedule coroutines from within synchronous functions. It returns the task it creates, but that return value is often ignored, which is a problem because [the event loop only keeps weak references to tasks](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) (the same goes for tasks created using `asyncio.ensure_future()`).
Unless I am missing something, this means that tasks created using `SSHConnection.create_task()` might be garbage collected and disappear at any time. The fact this has not yet caused any problems for anybody could just be because of some CPython implementation detail which asyncssh should not rely on.
### Proposed fix
- add a `_tasks` attribute to SSHConnections which is initialized to be an empty set
- `SSHConnection.create_task()` adds the task it creates to the set
- `SSHConnection._reap_task()` removes the task from the set

Even if the problem described above seems to never have occurred, the fix is really simple and should not impact performance, so maybe it should be done anyways just to be sure.

Birnendampf:
closes #716 
As discussed, `SSHConnection` now has a `_tasks` attribute. Created tasks are stored in it and removed when they are done.
When `wait_closed` is called, all running tasks are awaited.
What i am unsure about: should aborting the connection cancel all running tasks?  The idea here is that there needs to be a way to stop tasks that never finish. The user can try to close the connection gracefully, wait for it to close with a timeout and abort the connection if the timeout is exceeded.
The `abort()` method itself seems to be the right place to cancel all tasks, as `_force_close()` and `_cleanup()` are also called in a normal close/disconnect operation, so cancelling all tasks there would make awaiting them pointless

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -926,6 +926,8 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
 
         self._close_event = asyncio.Event()
 
+        self._tasks: Set[asyncio.Task[None]] = set()
+
         self._server_host_key_algs: Optional[Sequence[bytes]] = None
 
         self._logger = logger.get_child(
@@ -1083,6 +1085,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
                    task: 'asyncio.Task[None]') -> None:
         """Collect result of an async task, reporting errors"""
 
+        self._tasks.discard(task)
         # pylint: disable=broad-except
         try:
             task.result()
@@ -1101,6 +1104,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
 
         task = asyncio.ensure_future(coro)
         task.add_done_callback(partial(self._reap_task, task_logger))
+        self._tasks.add(task)
         return task
 
     def is_client(self) -> bool:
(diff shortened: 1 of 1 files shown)


ronf:
Keep in mind that `wait_closed()` can be called even when the connection is still fully open and in the middle of doing things that might take arbitrarily long to complete. The idea here is that something else in the application would automatically close the connection at some point, and that would cause `wait_closed()` to unblock. So, while an application is free to wrap a timeout around `wait_closed()` or otherwise decide to explicitly abort a connection on something like a timer, I don't think SSHConnection should ever do such a thing automatically.

Related to that, the cleanup functions in AsyncSSH are already designed to unblock everything related to network I/O on a closed or aborted connection. There's also already code to automatically cancel any auth in progress, as well as canceling the login and keepalive timers.

So, I'm wondering if the best option in `wait_closed()` might be to leave that as-is.  After all, it might not even be called, so there's no guarantee the tasks will finish before the SSHConnection is abandoned. Similarly, we can't count on `abort()` being called (in fact, it's relatively uncommon). The common function to do this in would be `_cleanup()`, but I think it's generally better to let tasks finish cleanly rather than cancelling them. If they hang around long enough for the garbage collector to cancel them after the strong reference in SSHConnection has gone away, that actually seems like it might be the right thing.

Another factor here is that `_cleanup()` and related functions like `_force_close()` are not async, so they can't actually wait on anything.

PR #720, opened 2024-11-22
Match - allow negation

commonism:
e.g.
Match host *.example.org !host jump.example.org

### DIFF
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -153,8 +153,15 @@ class SSHConfig:
 
         # pylint: disable=unused-argument
 
+        negate = False
+
         while args:
             match = args.pop(0).lower()
+
+            if match[0] == "!":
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for this!

As mentioned in #721, I have some other outstanding changes in config.py at the moment, but once that's committed I'll look at what it will take to pull this in.

ronf:
Thanks again for contributing this code! I wasn't able to merge it directly because of some pretty significant changes as a result of the hostname canonicalization work. However, I ended up incorporating an adapted version of these changes into commit b99e8d5 in the "develop" branch, alongside adding support for the "canonical" and "final" match keywords.

commonism:
Supporting match final ! is a common pattern in ProxyJump configurations, really great to see this supported.
Thanks for your effort.

PR #721, opened 2024-11-22
Match final

commonism:
adding support for Match final to SSHClientConfig

### DIFF
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -31,6 +31,7 @@ from pathlib import Path, PurePath
 from subprocess import DEVNULL
 from typing import Callable, Dict, List, NoReturn, Optional, Sequence
 from typing import Set, Tuple, Union, cast
+from xml.etree.ElementTree import canonicalize
 
 from .constants import DEFAULT_PORT
 from .logging import logger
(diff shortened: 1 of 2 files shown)


commonism:
Hi,

I've had the pleasure with Match final [already](https://github.com/paramiko/paramiko/pull/1907) and it just came back.

As I'm new to asyncssh, I'd appreciate if you could have a look and guide me when required.

ronf:
I appreciate your effort here, but as it happens I actually implemented support for SSH hostname canonicalization last weekend and part of that change included adding support for the "canonical" and "final" gestures in the config parser. The code isn't checked in yet, but if all goes well I should be able to commit a first cut of these changes by tomorrow.

Note that support for "final" isn't just a matter of running the parser twice. There can be other differences between the runs, such as the hostname canonicalization that needs to be reflected between the two passes. Also, when canonicalization occurs, both "canonical" and "final" should be handled in the same pass. When canonicalization doesn't occur, the second pass isn't actually needed at all unless "Match final" is present in the config. So, I actually look for that in the first pass and skip the second pass when neither applies.

commonism:
As hostname canonicalization is heavy lifting (CNAMEs), I choose to start of with final as warm up. 
Thanks for asyncssh, looking into it is time spent well.

As a side note, using PR's to write to the development branch and have it protected from direct pushes gives some points in software quality metric tooling.

PR #748, opened 2025-04-16
Add missing return type annotation

natehardison:


ronf:
Thanks for catching this!

PR #749, opened 2025-04-17
Add missing Awaitable return types

natehardison:
`SSHServer.*_requested()` callbacks can return Awaitables of their respective Session types, or a tuple consisting of a channel and a Session awaitable.

### DIFF
--- a/asyncssh/server.py
+++ b/asyncssh/server.py
@@ -20,7 +20,7 @@
 
 """SSH server protocol handler"""
 
-from typing import TYPE_CHECKING, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


ronf:
Thanks for submitting this!

You're right that all of the places you identified can accept a coroutine. However, I have a helper type called MaybeAwait that I generally use for signaling this. It wraps around a return type and automatically adds both the plain return type and an Awaitable of that type, cutting down on some of the duplication which would otherwise have been required here.

See commit 266a25f for the MaybeAwait version of this change. It should be the equivalent of what you suggested. If you find any problem with it, please let me know!

PR #752, opened 2025-05-04
Allow more flexibility in specifing env values

boris-42:
Instead of checking is it instance of dict check that it has "items" attribute. 

This way will work with any dict like object


ronf:
I ran into some typing errors in mypy when I switched over to the hasattr() version of this fix. So, I had to make additional changes (see commit 71f9b8a). The runtime behavior should be the same, though. Let me know if you have any problems with this version.

boris-42:
Sure, thanks!

Issue #777, opened 2025-10-23
SSH-1.99 SSH Identification String from Client


Jayson-Fong:
Running an AsyncSSH SSH server currently does not accept SSH identification strings prefixed with `SSH-1.99-`.

https://github.com/ronf/asyncssh/blob/d2812902357a64d5dc9ce8fe71b6ce3bc1b159e7/asyncssh/connection.py#L1553-L1554

When a client sends such an identification string, the server forcibly closes the connection.

https://github.com/ronf/asyncssh/blob/d2812902357a64d5dc9ce8fe71b6ce3bc1b159e7/asyncssh/connection.py#L1580

While this is not specified in an RFC, and rather [RFC 4253](https://datatracker.ietf.org/doc/html/rfc4253#section-5.1) specifies the protoversion for the server rather than a client, many legacy clients leverage a `1.99` protoversion to indicate that they support both SSHv1 and SSHv2.

SSHv1.5 is the final revision for SSHv1, so supporting this should generally not cause issues with incorrectly _failing_ to close the connection. We can observe such support in other SSH server libraries such as [paramiko](https://github.com/paramiko/paramiko/blob/8697158113a3eab25ed1d1d541ebe2cad10169a7/paramiko/transport.py#L2390-L2392).

ronf:
Fixed via #778 - thanks for the PR! This is now available in the "develop" branch and will be included in the next release.

PR #778, opened 2025-10-23
Support for SSH-1.99 SSH Identification String from Client

Jayson-Fong:
This change addresses #777 by dropping the `self.is_client()` condition when checking for `SSH-1.99` in the protoversion, thereby allowing it to apply to both client and server instances.

https://github.com/ronf/asyncssh/blob/d2812902357a64d5dc9ce8fe71b6ce3bc1b159e7/asyncssh/connection.py#L1553-L1554

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -1550,8 +1550,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
 
         self._inpbuf = self._inpbuf[idx+1:]
 
-        if (version.startswith(b'SSH-2.0-') or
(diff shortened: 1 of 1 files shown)


ronf:
Given that [RFC 4253](https://datatracker.ietf.org/doc/html/rfc4253#section-5.1) is nearly 20 years old at this point, I don't know how much value there is in continuing to support SSH 1.99 as an identification string. However, I don't see any real harm in it, so I've gone ahead and merged this change in the AsyncSSH "develop" branch. Thanks for your PR!

PR #786, opened 2025-11-25
Fix pre-auth banner decoding failure on invalid UTF-8 characters

Le-Syl21:
The client-side processing of SSH userauth banner messages was using strict UTF-8 decoding (the Python default), which caused a ProtocolError to be raised whenever the server sent a banner containing non-UTF-8 bytes. This happened even when users specified encoding='utf-8' and errors='replace' in their connection options, as these parameters were not being applied to the internal banner decoding logic.

This change adds errors='replace' to both the UTF-8 decoding of the banner message and the ASCII decoding of the language tag. Invalid bytes will now be replaced with the Unicode replacement character (U+FFFD) instead of causing the connection to fail.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -2638,8 +2638,8 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         packet.check_end()
 
         try:
-            msg = msg_bytes.decode('utf-8')
-            lang = lang_bytes.decode('ascii')
+            msg = msg_bytes.decode('utf-8', errors='replace')
+            lang = lang_bytes.decode('ascii', errors='replace')
         except UnicodeDecodeError:
             raise ProtocolError('Invalid userauth banner') from None
 


ronf:
Technically, sending invalid UTF-8 data is a protocol error, according to [RFC 4252 section 5.4](https://datatracker.ietf.org/doc/html/rfc4252#section-5.4):

```
5.4.  Banner Message

   In some jurisdictions, sending a warning message before
   authentication may be relevant for getting legal protection.  Many
   UNIX machines, for example, normally display text from /etc/issue,
   use TCP wrappers, or similar software to display a banner before
   issuing a login prompt.

   The SSH server may send an SSH_MSG_USERAUTH_BANNER message
   at any time after this authentication protocol starts and before
   authentication is successful.  This message contains text to be
   displayed to the client user before authentication is attempted.  The
   format is as follows:

      byte      SSH_MSG_USERAUTH_BANNER
      string    message in ISO-10646 UTF-8 encoding [RFC3629]
      string    language tag [RFC3066]
```

By default, I think AsyncSSH should continue to follow the RFC and treat this as a protocol error. This really seems like something that should be fixed on the misbehaving server. I could see potentially adding a config option to relax this, but in that case it seems like it should apply to more than just the pre-auth banner. There are places where "reason" strings are sent, for example. That said, there are other places where I don't think I'd relax this even with a config knob, such as when sending usernames, hostnames, or pathnames, as the replacement character in the middle of such a string is unlikely to work properly.

It looks like the potential candidates for this would be:

- Disconnect reason
- Debug message
- Userauth banner
- Channel open failure reason

There are also some strings used in password and keyboard-interactive auth that might also benefit from this, but that complicates the implementation, as I'd need to make this setting available outside of the `SSHConnection` objects. So, I'm tempted to leave those out unless there's a specific request for it.

If I do relax this, I'm also thinking that 'backslashreplace' might be better than 'replace', as it would let the client see what the original byte values were that weren't valid.

As for `lang`, I'm thinking that leaving that as only allowing ASCII as dictated in the RFC might be best. Most SSH implementations I've looked at don't even bother to send a value for `lang`, and it's unlikely to include non-ASCII characters when it is provided.

Le-Syl21:
Thanks for the detailed explanation and for considering this.

I fully agree with using `backslashreplace` instead of `replace` — it makes more sense for debugging purposes.

To give you some context on my use case: I'm in a country where pre-auth banners have legal significance, and I'm collecting data from a large fleet of machines, including AIX systems that send invalid UTF-8 in their banners. I have connection rights but no authority to modify the banners, and reaching out to dozens of different system owners to fix their banners isn't realistic.

Currently, I'm falling back to Paramiko (which handles this gracefully) when AsyncSSH fails, but this failover adds significant overhead and slows down my collection process considerably.

While I understand the appeal of a broader `utf8_errors` option covering multiple message types, would a simpler, more targeted approach work as a first step? Something like `banner_errors='backslashreplace'` (defaulting to `'strict'`) that only affects pre-auth banner decoding. This keeps the scope minimal and the RFC-compliant behavior intact everywhere else.

If there's demand later for relaxing other message types, it could always be extended.

ronf:
Thanks for explaining your use case here in more detail. I'd be happy to help support this so you don't have to keep around the fallback code.

As for how broad to make this, the bulk of the changes are going to be related to propagating a new config option for this, and that work is the same whether we do this for just banner errors or for a broader set of messages. Also, the broader solution avoids future upgrade/downgrade issues if the option gets renamed.

I'll try to put something together in the next day or two for you to try.

ronf:
This ended up being easier than I expected. See below for a patch which adds a new `utf8_decode_errors` config argument when creating an SSH connection, defaulting to 'strict' as we discussed. This implements this new argument for the four cases listed above.

```patch
index 82a5ec1..86a2f55 100644
--- a/asyncssh/connection.py
+++ b/asyncssh/connection.py
@@ -877,6 +877,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         self._peer_addr = ''
         self._peer_port = 0
         self._tcp_keepalive = options.tcp_keepalive
+        self._utf8_decode_errors = options.utf8_decode_errors
         self._owner: Optional[Union[SSHClient, SSHServer]] = None
         self._extra: Dict[str, object] = {}

@@ -1042,6 +1043,11 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):

         return self._logger

+    def _decode_utf8(self, msg_bytes) -> str:
+        """Decode UTF-8 bytes, honoring utf8_decode_errors setting"""
+
+        return msg_bytes.decode('utf-8', self._utf8_decode_errors)
+
     def _cleanup(self, exc: Optional[Exception]) -> None:
         """Clean up this connection"""

@@ -2193,7 +2199,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         packet.check_end()

         try:
-            reason = reason_bytes.decode('utf-8')
+            reason = self._decode_utf8(reason_bytes)
             lang = lang_bytes.decode('ascii')
         except UnicodeDecodeError:
             raise ProtocolError('Invalid disconnect message') from None
@@ -2236,7 +2242,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         packet.check_end()

         try:
-            msg = msg_bytes.decode('utf-8')
+            msg = self._decode_utf8(msg_bytes)
             lang = lang_bytes.decode('ascii')
         except UnicodeDecodeError:
             raise ProtocolError('Invalid debug message') from None
@@ -2638,7 +2644,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         packet.check_end()

         try:
-            msg = msg_bytes.decode('utf-8')
+            msg = self._decode_utf8(msg_bytes)
             lang = lang_bytes.decode('ascii')
         except UnicodeDecodeError:
             raise ProtocolError('Invalid userauth banner') from None
@@ -2755,7 +2761,7 @@ class SSHConnection(SSHPacketHandler, asyncio.Protocol):
         packet.check_end()

         try:
-            reason = reason_bytes.decode('utf-8')
+            reason = self._decode_utf8(reason_bytes)
             lang = lang_bytes.decode('ascii')
         except UnicodeDecodeError:
             raise ProtocolError('Invalid channel open failure') from None
@@ -7278,6 +7284,7 @@ class SSHConnectionOptions(Options, Generic[_Options]):
     family: int
     local_addr: HostPort
     tcp_keepalive: bool
+    utf8_decode_errors: str
     canonicalize_hostname: Union[bool, str]
     canonical_domains: Sequence[str]
     canonicalize_fallback_local: bool
@@ -7323,6 +7330,7 @@ class SSHConnectionOptions(Options, Generic[_Options]):
                 passphrase: Optional[BytesOrStr],
                 proxy_command: DefTuple[_ProxyCommand], family: DefTuple[int],
                 local_addr: DefTuple[HostPort], tcp_keepalive: DefTuple[bool],
+                utf8_decode_errors: str,
                 canonicalize_hostname: DefTuple[Union[bool, str]],
                 canonical_domains: DefTuple[Sequence[str]],
                 canonicalize_fallback_local: DefTuple[bool],
@@ -7387,6 +7395,8 @@ class SSHConnectionOptions(Options, Generic[_Options]):
         self.tcp_keepalive = cast(bool, tcp_keepalive if tcp_keepalive != ()
             else config.get('TCPKeepAlive', True))

+        self.utf8_decode_errors = utf8_decode_errors
+
         self.canonicalize_hostname = \
             cast(Union[bool, str], canonicalize_hostname
                  if canonicalize_hostname != ()
@@ -7812,6 +7822,13 @@ class SSHClientConnectionOptions(SSHConnectionOptions):
        :param tcp_keepalive: (optional)
            Whether or not to enable keepalive probes at the TCP level to
            detect broken connections, defaulting to `True`.
+       :param utf8_decode_errors: (optional)
+           Error handling strategy to apply when UTF-8 decode errors
+           occur in SSH protocol messages, defaulting to 'strict'
+           which shuts down the connection with a ProtocolError.
+           Choosing other strategies can allow the message parsing
+           to proceed with invalid bytes in the message being removed
+           or replaced.
        :param canonicalize_hostname: (optional)
            Whether or not to enable hostname canonicalization, defaulting
            to `False`, in which case hostnames are passed as-is to the
@@ -7984,6 +8001,7 @@ class SSHClientConnectionOptions(SSHConnectionOptions):
        :type keepalive_interval: *see* :ref:`SpecifyingTimeIntervals`
        :type keepalive_count_max: `int`
        :type tcp_keepalive: `bool`
+       :type utf8_decode_errors: `str`
        :type canonicalize_hostname: `bool` or `'always'`
        :type canonical_domains: `list` of `str`
        :type canonicalize_fallback_local: `bool`
@@ -8069,6 +8087,7 @@ class SSHClientConnectionOptions(SSHConnectionOptions):
                 family: DefTuple[int] = (),
                 local_addr: DefTuple[HostPort] = (),
                 tcp_keepalive: DefTuple[bool] = (),
+                utf8_decode_errors: str = 'strict',
                 canonicalize_hostname: DefTuple[Union[bool, str]] = (),
                 canonical_domains: DefTuple[Sequence[str]] = (),
                 canonicalize_fallback_local: DefTuple[bool] = (),
@@ -8180,10 +8199,11 @@ class SSHClientConnectionOptions(SSHConnectionOptions):

         super().prepare(config, client_factory or SSHClient, client_version,
                         host, port, tunnel, passphrase, proxy_command, family,
-                        local_addr, tcp_keepalive, canonicalize_hostname,
-                        canonical_domains, canonicalize_fallback_local,
-                        canonicalize_max_dots, canonicalize_permitted_cnames,
-                        kex_algs, encryption_algs, mac_algs, compression_algs,
+                        local_addr, tcp_keepalive, utf8_decode_errors,
+                        canonicalize_hostname, canonical_domains,
+                        canonicalize_fallback_local, canonicalize_max_dots,
+                        canonicalize_permitted_cnames, kex_algs,
+                        encryption_algs, mac_algs, compression_algs,
                         signature_algs, host_based_auth, public_key_auth,
                         kbdint_auth, password_auth, x509_trusted_certs,
                         x509_trusted_cert_paths, x509_purposes, rekey_bytes,
@@ -8636,6 +8656,13 @@ class SSHServerConnectionOptions(SSHConnectionOptions):
        :param tcp_keepalive: (optional)
            Whether or not to enable keepalive probes at the TCP level to
            detect broken connections, defaulting to `True`.
+       :param utf8_decode_errors: (optional)
+           Error handling strategy to apply when UTF-8 decode errors
+           occur in SSH protocol messages, defaulting to 'strict'
+           which shuts down the connection with a ProtocolError.
+           Choosing other strategies can allow the message parsing
+           to proceed with invalid bytes in the message being removed
+           or replaced.
        :param canonicalize_hostname: (optional)
            Whether or not to enable hostname canonicalization, defaulting
            to `False`, in which case hostnames are passed as-is to the
@@ -8732,6 +8759,7 @@ class SSHServerConnectionOptions(SSHConnectionOptions):
        :type keepalive_interval: *see* :ref:`SpecifyingTimeIntervals`
        :type keepalive_count_max: `int`
        :type tcp_keepalive: `bool`
+       :type utf8_decode_errors: `str`
        :type canonicalize_hostname: `bool` or `'always'`
        :type canonical_domains: `list` of `str`
        :type canonicalize_fallback_local: `bool`
@@ -8790,6 +8818,7 @@ class SSHServerConnectionOptions(SSHConnectionOptions):
                 family: DefTuple[int] = (),
                 local_addr: DefTuple[HostPort] = (),
                 tcp_keepalive: DefTuple[bool] = (),
+                utf8_decode_errors: str = 'strict',
                 canonicalize_hostname: DefTuple[Union[bool, str]] = (),
                 canonical_domains: DefTuple[Sequence[str]] = (),
                 canonicalize_fallback_local: DefTuple[bool] = (),
@@ -8865,10 +8894,11 @@ class SSHServerConnectionOptions(SSHConnectionOptions):

         super().prepare(config, server_factory or SSHServer, server_version,
                         host, port, tunnel, passphrase, proxy_command, family,
-                        local_addr, tcp_keepalive, canonicalize_hostname,
-                        canonical_domains, canonicalize_fallback_local,
-                        canonicalize_max_dots, canonicalize_permitted_cnames,
-                        kex_algs, encryption_algs, mac_algs, compression_algs,
+                        local_addr, tcp_keepalive, utf8_decode_errors,
+                        canonicalize_hostname, canonical_domains,
+                        canonicalize_fallback_local, canonicalize_max_dots,
+                        canonicalize_permitted_cnames, kex_algs,
+                        encryption_algs, mac_algs, compression_algs,
                         signature_algs, host_based_auth, public_key_auth,
                         kbdint_auth, password_auth, x509_trusted_certs,
                         x509_trusted_cert_paths, x509_purposes,
```

ronf:
Commit 254d735 in the "develop" branch now includes an expanded version of this above. It now covers channel exit signal reasons and SFTPError reasons in addition to the four items listed above. I also added unit tests for this new functionality. Let me know if you have any issues with it.

Le-Syl21:
Hi,
For a BAD banner_auth server:
BEFORE FIX:

SSH: • TOTO (AIX): Invalid userauth banner
SCP: • TOTO: PUSH failed: Erreur SFTP PUT sur TOTO: Invalid userauth banner

AFTER FIX (utf8_decode_errors=replace):

SSH: ✅ Terminé (1): TOTO
SCP: ✅ Fichier uploadé avec succès (permissions: 0o640)

AFTER FIX (utf8_decode_errors=backslashreplace):

SSH: ✅ Terminé (1): TOTO
SCP: ✅ Fichier uploadé avec succès (permissions: 0o640)

I haven't seen any change for GOOD banner_auth servers tested on:

AIX 7
Linux RHEL7 → 9

Sorry, no more Solaris and HP-UX :-p
Good to merge! Thanks for your help and the quick fix!

ronf:
Great - thanks for the confirmation! This will be included in the next release. In the meantime, feel free to use the version in the "develop" branch.

Le-Syl21:
Hello,

Would it be possible to quickly include this fix in a new version?

I need to deliver the final version of my code and my client only has access to pip through a Nexus repository. Their servers don't have internet access, so they can't reach the development GitHub repository either.

I would like to be able to add a new version to my requirements and have it work out of the box.

Thanks in advance

PR #788, opened 2026-01-03
use context managers with aiofiles open

gsauthof:
to make the tests more explicit and eliminate possible race conditions

---

This pull-request is triggered by a test case failure on Fedora Rawhide, i.e. when packaging asyncssh 2.22.0:

```
Test validation of X.509 self-signed certificate ... ok
======================================================================
FAIL: test_stdout_aiofile (tests.test_process._TestAsyncFileRedirection.test_stdout_aiofile)
Test with stdout redirected to an aiofile
----------------------------------------------------------------------
  File "/builddir/build/BUILD/python-asyncssh-2.22.0-build/asyncssh-2.22.0/tests/util.py", line 92, in async_wrapper
    return self.loop.run_until_complete(coro(self, *args, **kwargs))
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/builddir/build/BUILD/python-asyncssh-2.22.0-build/asyncssh-2.22.0/tests/test_process.py", line 1252, in test_stdout_aiofile
    self.assertEqual(stdout_data, data)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
AssertionError: '' != '140735626508016'
+ 140735626508016
----------------------------------------------------------------------
Ran 1727 tests in 327.287s
FAILED (failures=1, skipped=26)
```
https://kojipkgs.fedoraproject.org//work/tasks/7459/140687459/build.log
https://koji.fedoraproject.org/koji/taskinfo?taskID=140687459

I couldn't reproduce the issue locally, on Fedora 43, although the aiofiles version is the same:

```
python3-aiofiles-25.1.0-2.fc44.noarch
```
https://kojipkgs.fedoraproject.org//work/tasks/7459/140687459/root.log

Looking at the code, invoking `aiofiles.open` from a context manager arguably is more idiomatic and explicit, anyways, thus, the code change seems worthwhile, in any case.

I tested the change with the fedora rawhide build systems and all tests succeed:

- https://koji.fedoraproject.org/koji/taskinfo?taskID=140690280 (just the affected test case patched)
- https://koji.fedoraproject.org/koji/taskinfo?taskID=140691830 (full patch from this pull-request)
- https://koji.fedoraproject.org/koji/taskinfo?taskID=140692451 (as before, but forcing ppc64le target, as in the failing run)

I'm not familiar with the internals of aiofiles, but since it uses a thread pool internally, a race condition is plausible.
The aiofiles issue tracker has a few tickets about people having issues invoken open outside of a context manager, which might or might not be related.

Also, CI systems often are more loaded than developer systems, thus a failure may be more likely in such an environment.

NB: For portable python packages, the fedora build system select a host at random, i.e. the binary package is built at random on aarch64, x86_64, ppc64le etc. systems.

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -1210,10 +1210,9 @@ class _TestAsyncFileRedirection(_TestProcess):
         with open('stdin', 'w') as file:
             file.write(data)
 
-        file = await aiofiles.open('stdin', 'r')
-
-        async with self.connect() as conn:
-            result = await conn.run('echo', stdin=file)
+        async with aiofiles.open('stdin', 'r') as file:
+            async with self.connect() as conn:
+                result = await conn.run('echo', stdin=file)
 
         self.assertEqual(result.stdout, data)
         self.assertEqual(result.stderr, data)
@@ -1227,10 +1226,9 @@ class _TestAsyncFileRedirection(_TestProcess):
         with open('stdin', 'wb') as file:
             file.write(data)
 
-        file = await aiofiles.open('stdin', 'rb')
-
-        async with self.connect() as conn:
-            result = await conn.run('echo', stdin=file, encoding=None)
+        async with aiofiles.open('stdin', 'rb') as file:
+            async with self.connect() as conn:
+                result = await conn.run('echo', stdin=file, encoding=None)
 
         self.assertEqual(result.stdout, data)
         self.assertEqual(result.stderr, data)
@@ -1241,10 +1239,9 @@ class _TestAsyncFileRedirection(_TestProcess):
 
         data = str(id(self))
 
-        file = await aiofiles.open('stdout', 'w')
-
-        async with self.connect() as conn:
-            result = await conn.run('echo', input=data, stdout=file)
+        async with aiofiles.open('stdout', 'w') as file:
+            async with self.connect() as conn:
+                result = await conn.run('echo', input=data, stdout=file)
 
         with open('stdout') as file:
             stdout_data = file.read()
@@ -1275,11 +1272,10 @@ class _TestAsyncFileRedirection(_TestProcess):
 
         data = str(id(self)).encode() + b'\xff'
 
-        file = await aiofiles.open('stdout', 'wb')
-
-        async with self.connect() as conn:
-            result = await conn.run('echo', input=data, stdout=file,
-                                    encoding=None)
+        async with aiofiles.open('stdout', 'wb') as file:
+            async with self.connect() as conn:
+                result = await conn.run('echo', input=data, stdout=file,
+                                        encoding=None)
 
         with open('stdout', 'rb') as file:
             stdout_data = file.read()
@@ -1297,11 +1293,10 @@ class _TestAsyncFileRedirection(_TestProcess):
         with open('stdin', 'w') as file:
             file.write(data)
 
-        file = await aiofiles.open('stdin', 'r')
-
-        async with self.connect() as conn:
-            result = await conn.run('delay', stdin=file,
-                                    stderr=asyncssh.DEVNULL)
+        async with aiofiles.open('stdin', 'r') as file:
+            async with self.connect() as conn:
+                result = await conn.run('delay', stdin=file,
+                                        stderr=asyncssh.DEVNULL)
 
         self.assertEqual(result.stdout, data)
 


ronf:
Thanks for reporting this! I have seen race conditions in the past with aiofiles, but not for a long time and not this specific issue, even when running it through GitHub automated workflows.

Looking at your change, it appears to be doing two things. First, I think it is now opening the SSH connection before it opens the aiofile, and second it is now closing the aiofile (due to the context manager) at the end. The latter doesn't seem like it should matter to the current test, but there's a possibility that leaving that file open (cleaned up later by garbage collection) could potentially cause problems for future tests, and I agree it's better to reliably clean up before each test exits.

If you have a way to reproduce this, would you mind switching the order of "conn" and "file" in the "async with" and see whether that makes a difference? While both orders should be equally good, I'm just curious which of the two changes mattered here. If the ordering doesn't matter, I would lean toward preserving the existing ordering (aiofile open before opening up conn).

Also, minor nit: the new code now has some 80-character lines. I'd suggest breaking it into two separate "async with" lines, indenting the "result =" line inside. Basically, just convert the "file =" into an "async with" and leave the rest of the code unchanged (except for indents).

gsauthof:
Ok, I've triggered another build run, with the original construction order.

Regarding the syntax: Another alternative is to use parentheses, i.e. instead of:

```
        async with aiofiles.open('stdout', 'w') as file:
            async with self.connect() as conn:
                result = await conn.run('echo', input=data, stdout=file)
```

do:

```
        async with (aiofiles.open('stdout', 'w') as file,
                    self.connect() as conn):
            result = await conn.run('echo', input=data, stdout=file)
```

Advantage: limits indentation
Disadvantages: more parentheses

I'm fine with either style

gsauthof:
The build job is through and still succeeds:

https://koji.fedoraproject.org/koji/taskinfo?taskID=140696213

(i.e. `aiofile.open()` still in context manager, but in the original order, again forced on ppc64le)

---

For comparison I repeated the vanilla build (without any patches) under ppc64le:
This time another test case fails, besides the original one, in a similar way:

```
Test validation of X.509 self-signed certificate ... ok
======================================================================
FAIL: test_stdout_aiofile (tests.test_process._TestAsyncFileRedirection.test_stdout_aiofile)
Test with stdout redirected to an aiofile
----------------------------------------------------------------------
  File "/builddir/build/BUILD/python-asyncssh-2.22.0-build/asyncssh-2.22.0/tests/util.py", line 92, in async_wrapper
    return self.loop.run_until_complete(coro(self, *args, **kwargs))
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/builddir/build/BUILD/python-asyncssh-2.22.0-build/asyncssh-2.22.0/tests/test_process.py", line 1252, in test_stdout_aiofile
    self.assertEqual(stdout_data, data)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
AssertionError: '' != '140735821870832'
+ 140735821870832
======================================================================
FAIL: test_stdout_binary_aiofile (tests.test_process._TestAsyncFileRedirection.test_stdout_binary_aiofile)
Test with stdout redirected to an aiofile in binary mode
----------------------------------------------------------------------
  File "/builddir/build/BUILD/python-asyncssh-2.22.0-build/asyncssh-2.22.0/tests/util.py", line 92, in async_wrapper
    return self.loop.run_until_complete(coro(self, *args, **kwargs))
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/builddir/build/BUILD/python-asyncssh-2.22.0-build/asyncssh-2.22.0/tests/test_process.py", line 1287, in test_stdout_binary_aiofile
    self.assertEqual(stdout_data, data)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
AssertionError: b'' != b'140735819706672\xff'
----------------------------------------------------------------------
Ran 1727 tests in 339.092s
FAILED (failures=2, skipped=26)
```
https://kojipkgs.fedoraproject.org//work/tasks/6798/140696798/build.log
https://koji.fedoraproject.org/koji/taskinfo?taskID=140696798

ronf:
Thanks for re-testing this, and updating the PR. I'll merge this shortly. The results suggest that the problem is related to the failure to clean up the aiofile before the test exits and another is started, which is what I was sort of expecting to see.

I thought about using parens and/or a backslash line continuation, but given the body in this case is just a single short line, I'd lean toward two separate "async with" statements because that's more explicit about the order they are created. With the comma, there could be a question about which CM is created first.

PR #789, opened 2026-01-03
remove libnacl dependency declarations

gsauthof:
after release 2.22.0 elimnated code usage

cf. https://asyncssh.readthedocs.io/en/latest/changes.html#release-2-22-0-21-dec-2025


ronf:
Thanks for this! I also noticed some references to libnacl/libsodium in api.rst which I'll clean up in a separate commit.

PR #794, opened 2026-02-13
fix: ensure file handle is closed in update_xauth on exception

taovinci0:
## Problem
The `update_xauth` function in `x11.py` opens a file but doesn't properly close it if an exception occurs during write operations or `walk_xauth`. This can lead to resource leaks, especially in long-running SSH servers.

## Solution
Wrap the file operations in a try-finally block to ensure the file handle is always closed, even if an exception is raised.

## Impact
- Prevents file descriptor leaks in X11 forwarding
- Improves reliability for long-running SSH servers
- No functional changes to the logic

### DIFF
--- a/asyncssh/x11.py
+++ b/asyncssh/x11.py
@@ -491,19 +491,20 @@ async def update_xauth(auth_path: Optional[str], host: str, dpynum: str,
     if not new_file:
         raise ValueError('Unable to acquire Xauthority lock')
 
-    new_entry = SSHXAuthorityEntry(XAUTH_FAMILY_HOSTNAME, host,
-                                   dpynum, auth_proto, auth_data)
+    try:
+        new_entry = SSHXAuthorityEntry(XAUTH_FAMILY_HOSTNAME, host,
+                                       dpynum, auth_proto, auth_data)
 
-    new_file.write(bytes(new_entry))
+        new_file.write(bytes(new_entry))
 
-    for entry in walk_xauth(auth_path):
-        if (entry.family != new_entry.family or entry.addr != new_entry.addr or
-                entry.dpynum != new_entry.dpynum):
-            new_file.write(bytes(entry))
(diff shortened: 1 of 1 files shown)


ronf:
Thanks for the report and the suggested fix.

This is a good change which I'm happy to take. I'm curious, though -- in your experience, did this actually turn into a long-term file descriptor leak, or did the file descriptors get closed when the file objects were garbage collected? It's better to do this explicitly as you have here, but my understanding is that the file descriptors will be closed at the next garbage collection after the file object goes out of scope. So, in the end this probably wouldn't turn into a resource leak in a long-running server. The concern would be more about a short term issue of hitting the limit on open files if a caller ended up making a lot of calls to this code and kept experiencing an error.

One minor nit with the change: The extra level of indent is enough to make one of the lines wider than 80 characters. I would suggest adding another line break there:

```python
            if (entry.family != new_entry.family or
                    entry.addr != new_entry.addr or
                    entry.dpynum != new_entry.dpynum):
```

Also, I'm thinking it might be better to use a `with new_file:` block rather than the try..finally. Either one should fix the problem, but the "with" is a bit simpler.

taovinci0:
Thanks for the review! Updated with both suggestions - switched to `with` context manager and fixed the line length.

ronf:
Did you mean to close this?

ronf:
Thanks for the report and fix!

ronf:
As it turns out, this change doesn't work on Windows. In the previous code, new_file is closed BEFORE the call to os.replace() which moves that file on top of the previous instance of the file. Doing the close after os.replace() works ok on Linux and macOS, but not on Windows. For Windows, the file must be closed before trying to do the file rename.

This should be fixed by 1af489f. It runs ok on Windows in my local testing, but we'll have to see if the GitHub actions are ok with this.

PR #800, opened 2026-03-21
Fix NameError in PQDH when handling invalid algorithms

RinZ27:
Fixed a potential NameError in PQDH.__init__ when an invalid algorithm is provided.

The current implementation tries to use `oqs_name` inside the KeyError handler before it's actually assigned in the try block. This leads to a NameError that masks the intended ValueError about the unknown algorithm. 

Switched to using `alg_name.decode()` in the error message to ensure it's both accurate and helpful for debugging configuration issues. I also added a test case in `tests/test_kex.py` to verify the fix and prevent regressions.

### DIFF
--- a/asyncssh/crypto/pq.py
+++ b/asyncssh/crypto/pq.py
@@ -59,7 +59,7 @@ class PQDH:
             self.ciphertext_bytes, self.secret_bytes, \
             oqs_name = _pq_algs[alg_name]
         except KeyError: # pragma: no cover, other algs not registered
-            raise ValueError(f'Unknown PQ algorithm {oqs_name}') from None
+            raise ValueError(f'Unknown PQ algorithm {alg_name.decode()}') from None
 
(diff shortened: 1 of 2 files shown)


ronf:
Thanks for the report!

I agree this is a bug and I can clean this up, but I'm curious how you ran into this issue. The reason that particular block is marked as "pragma: no cover" is that it shouldn't be possible for that KeyError to be raised. Supported algorithms must be registered before they can be used, and without that registration AsyncSSH would never get far enough to create a `PQDH` object for the invalid algorithm, and thus never enter that KeyError block. The code is only there as protection against a bug in AsyncSSH where an algorithm is registered but the internal alg name used in the registration doesn't match any of the names in `_pq_algs` in crypto/pq.py. It shouldn't be possible to trigger this from application code.

ronf:
Thanks for this PR - the fix is now merged into the "develop" branch.

PR #807, opened 2026-05-21
feat(forward): SSHForwardTracker hierarchy + per-connection tracker_factory for local and remote forwards

AlexMKX:
## Summary

Adds an optional, observer-only hook API for watching the lifecycle of
forwarded connections (open / close / bytes in each direction), driven by a
per-connection `tracker_factory`. The motivating use case is idle-based
auto-shutdown of a forwarding daemon, but the same hooks support connection
counting and passive traffic metrics.

This reflects the design worked out in the discussion below (superseding the
original single `ForwardTracker` Protocol proposal).

## API

A small class hierarchy mirroring the existing forwarder classes:

- **`SSHForwardTracker`** — base class. No-op default hooks shared by all forward
  types:
  - `connection_lost(exc)` — the forwarded connection closed (`exc=None` on clean
    close). Fires **exactly once** per connection.
  - `forward_local_bytes(data)` — a block forwarded local → tunnel.
  - `forward_remote_bytes(data)` — a block forwarded tunnel → local.
- **`SSHPortForwardTracker(SSHForwardTracker)`** — when the listener is a TCP
  port. Adds `connection_made(forwarder, orig_host, orig_port)`.
- **`SSHPathForwardTracker(SSHForwardTracker)`** — when the listener is a UNIX
  domain socket. Adds `connection_made(forwarder)` (no address args).

A new `tracker_factory: Callable[[], SSHForwardTracker] | None = None` keyword is
added to the eight forwarding methods, invoked **once per accepted
connection**. The tracker type follows the **listener** endpoint, not the
destination:

| method | tracker type |
|---|---|
| `forward_local_port` | `SSHPortForwardTracker` |
| `forward_local_port_to_path` | `SSHPortForwardTracker` |
| `forward_local_path` | `SSHPathForwardTracker` |
| `forward_local_path_to_port` | `SSHPathForwardTracker` |
| `forward_remote_port` | `SSHPortForwardTracker` |
| `forward_remote_port_to_path` | `SSHPortForwardTracker` |
| `forward_remote_path` | `SSHPathForwardTracker` |
| `forward_remote_path_to_port` | `SSHPathForwardTracker` |

The byte hooks name the host where the data was **generated**, so for a remote
forward the mapping is mirrored: `forward_remote_bytes` sees what the client of
the remote listener sent, and `forward_local_bytes` sees what the local
destination sent back.

### Semantics (per discussion)

- **Observer-only.** Byte hooks return nothing and never alter the forwarded
  data; transform/parsing use cases should drop down to `create_connection()` /
  `open_connection()` instead.
- **No opt-in machinery for byte hooks** — rely on the parent's no-op defaults.
  Hooks fire once per SSH message (not per byte), behind SSH decryption, so the
  Python call cost is negligible in any realistic workload.
- **Single `connection_lost(exc)`**, symmetric with the single `connection_made`
  (no per-direction split).
- Every hook has a no-op default, so a subclass overrides only what it needs.
  Exceptions raised by a hook or by the factory are caught and discarded so a
  buggy tracker can never break forwarding.
- `connection_lost` is guarded so it fires exactly once even though
  `_forward()` notifies manually on a channel-open failure and the local
  transport's later close would otherwise notify again.

## Scope

Local **and** remote forwarding. `forward_socks` is still left for a follow-up;
the hierarchy + factory accept it without rework.

> **Note on review history:** this PR originally covered local forwarding only,
> and your 1 Aug "these changes all look good" applies to that state, up to
> `2bf4164`. Commits `bc19857..96624f4` add the remote support you suggested as
> the next step, so that part has not been reviewed yet. The local half is
> refactored — the tracker plumbing moved into a shared `SSHTrackedForwarder`
> base so the remote forwarders reuse it instead of duplicating it — but is
> intended to be behaviour-preserving, and the existing local tracker tests pass
> unchanged.

One lint-config change came with it: `connection.py` was 26 lines under
`max-module-lines`, so `pylintrc` goes 10000 → 11000. Happy to drop that and
take the C0302 instead if you would rather keep the config untouched.

Trackers on a remote forward are created when the listener connection is
accepted, so a connection whose local destination cannot be opened still
produces exactly one tracker, with `connection_made` followed by
`connection_lost(ChannelOpenError)`. That is the question raised in the thread
below — easy to flip if you would prefer a tracker only once both ends are up.

## Docs

A dedicated "Forward Tracker Classes" section in `api.rst` (factory pattern,
example, and the list of tracker-aware methods). Per your note, no `changes.rst`
entry is included.

## Tests

New tests in `tests/test_forward.py` covering both tracker subclasses,
`connection_made` / `connection_lost` ordering, per-connection factory
invocation, both byte-hook directions, "lost fires exactly once" on a denied
forward, and exception isolation for both a buggy factory and buggy hooks —
each mirrored for the remote methods, including the `*_to_path` / `*_to_port`
cross cases.

The byte-direction tests deliberately use a distinct request and response
rather than an echo server: an echo-based test passes even if the two byte
hooks are wired backwards, which was verified by mutation. There is also a
deterministic test that a cancelled destination open still delivers
`connection_lost`.


ronf:
Thanks for the PR. Could you say a little more about the use case you have in mind for this? I'm not really seeing how this would be able to do things like byte counts or idle tracking if you are only sending connection_made and connection_lost messages through the tracker.

I'm wondering if some of this could be handled by using the existing accept_handler. It only runs right now on a new client connection coming in, giving you the client IP address and port and even lets you decide based on that whether to allow the forwarded connection or not. It doesn't currently run when a connection closes (either cleanly or with an error), but that could be added as something like an error_handler. In the future, support for a progress_handler could also be added here to report on bytes transferred, much like what exists in AsyncSSH right now for the SFTP/SCP file copy functions.

AlexMKX:
Hello @ronf, thank you for the questions.

The main use case I have in mind is handling idle forwarded connections. This came from a practical need in this tool: https://github.com/AlexMKX/garuda-tunnel

It is basically a “poor man's Teleport” for automating Kubernetes access through Terraform or an Ansible Kubernetes inventory source. It runs as a daemon, exposes local ports through a bastion, and uses AsyncSSH under the hood.

For safety and operational visibility, I would like to track the number of live forwarded connections and apply an idle timeout to traffic going through the mapped ports. A per-byte counter would be better, of course, but for this use case an active connection count is a reasonable first approximation without inviting a whole committee of extra machinery into the room.

I agree that accept_handler is close to this area. The main gap is that it currently only covers the beginning of the connection. Having something symmetrical for connection close, perhaps an error_handler or close_handler, would probably solve a good part of this. A future progress_handler for byte counts would also be useful and would fit nicely with what AsyncSSH already does for SFTP/SCP transfers.

So I am not strongly attached to the exact tracker API shape from the PR. The important part for me is having a lightweight way to observe connection lifecycle events for forwarded connections, and later possibly traffic progress, without each user of port forwarding having to rebuild that logic around AsyncSSH internals.

ronf:
Thanks for the added context... I'm in the middle of some other feature work at the moment, but once that's wrapped up I'll take a closer look at this, probably in the next couple of weeks.

I'm thinking that you could still have a `ForwardTracker` class at the application level if you wanted here, and instantiate that with details of the connection it is associated with. The arguments to `accept_handler` and `close_handler` in `forward_local_port()` could point at the bound methods of that specific ForwardTracker object. The same could apply in the future for tracking byte counts.

In fact, you could probably have a factory method on `ForwardTracker` that actually called `forward_local_port()` with all the necessary bound methods as arguments. This will need to be an async method so you can't use the constructor directly, but this should work as an async class method if you wanted to avoid needing to call the constructor first yourself.

ronf:
I had a chance to get  back to this today.

My main concern about what we previously discussed is the lack of the ability to identify specific connections with this approach, especially if we want to extend it later to report specific byte counts being transferred on the connections. For local port forwarding (or even port to path), one could argue that orig_host and orig_port should uniquely identify a specific connection if you had different tracker instances per SSHListener instance, but there's no equivalent of that if we were to extend this to cover forwarding of UNIX domain socket connections. That's actually part of why I never implemented an accept handler for the `forward_local_path()` and `forward_local_path_to_port()` cases. There was nothing for the handler to use to decide which connections to allow and which to reject.

One possible way to address this would be to pass in a factory function when setting up forwarding. A new instance of this would be created per forwarded connection, and once forwarding is successfully set up for a connection, the factory function would be called to create a new tracking instance for that connection, and connection_made (and later connection_lost) would be called on these instances.

With this approach, other forms of monitoring should be pretty easy to add. For instance, there could be methods for data_sent() and data_received(), providing the actual data being forwarded, not just byte counts. I'm also thinking the call to connection_made() could possibly provide an `SSHForwarder` instance, allowing this to be used later to do something like forcibly close a specific connection from the tracker.

AlexMKX:
Factory-per-connection makes sense — it solves the UNIX-socket identity gap and scales cleanly to byte counts and force-close, which the aggregate observer can't. For my idle use case I'd just have the factory bump a shared counter and return a per-connection object whose connection_lost decrements it.

Before I rework the PR: could you pin the intended shape? Specifically — the parameter name on forward_local_port (tracker_factory?), whether the factory is sync or async, and what connection_made receives (an SSHForwarder, plus orig addr?). I'll adapt the PR (and the docs/tests) to match once the signature is settled.

ronf:
Yes - `tracker_factory` seems like a good name for the new argument, and it should default to `None` as in your original proposal here. This factory function should be synchronous. While there are places where callbacks can be either synchronous or asynchronous, the methods being mirrored here in things like SSHConnection SSHSession are generally synchronous, so I'd go with that as the initial plan here unless you can think of a need that one of these callbacks would need to be async.

To handle both TCP and UNIX domain socket cases, I think we may need two different protocols, though there could be a common parent for most of the methods. The difference would be in either `connection_made()` or the constructor to take arguments for `orig_host` and `orig_port` in the TCP case but no additional arguments in the UNIX domain socket case. I'm thinking probably extra arguments to `connection_made()` after the SSHForwarder instance, as you said.

In terms of naming on the tracker class, I would have `SSHForwardTracker` as a parent class with `SSHPortForwardTracker` and `SSHPathForwardTracker` as subclasses, mirroring the existing `SSHPortForwarder` and `SSHPathForwarder` classes.

We should be able to share a tracker for both `forward_local_port()` and `forward_local_port_to_path()`, and similarly share the other tracker for `forward_local_path()` and `forward_local_path_to_port()`. I haven't had a chance to think much about whether we'd also want this trackers to work for the equivalent `forward_remote` methods, but at first thought it seems like we should be able to make all four of those methods work as well here, using the same trackers for both the local & remote cases.

We could also look into supporting `forward_socks()` in the future, but I think that's something which can wait until we have the base cases working. I could even see splitting up the work to start with only local forwarding with remote coming later, as long as we fill in enough of the design to know it CAN be added later without major re-work.

In terms if the callback methods, I'm thinking:

- `connection_made(forwarder: SSHForwarder, orig_host: str, orig_port: int)` for TCP or `connection_made(forwarder: SSHForwarder)` for UNIX domain sockets when a new tunneled connection is opened.
- `connection_lost(exc: Optional[Exception])` for reporting a clean close or an error on a tunneled connection.
- `forward_local_bytes(data: bytes)` for forwarding bytes from a local socket to the SSH connection
- `forward_remote_bytes(data: bytes)` for forwarding remote bytes from the SSH connection to a local socket
- `close()` to forcibly close the tunnel.

I'm thinking for now we probably don't need to separately track EOF for tunneled connections. While it's possible for half-open connections to exist, it's not common, especially for forwarding. There MIGHT be value in reporting the optional exception differently for the two directions, since technically we could get a clean close on one side but an exception on the other, or two different exceptions for the two sides. I don't know how useful this would be, but if we did this, I'm thinking `connection_lost()` would get changed to `local_connection_lost()` and `remote_connection_lost()`. Generally, these calls would happen in quick succession, but the ordering of the two would reflect  which side closed first (in addition to letting us report different exceptions).

ronf:
Oh - one other thing. I would not include the changes.rst file in the PR. A new changes.rst entry would only appear when a new release is cut, and would be a much less detailed than what was proposed here. That said, we could add some documentation to the new SSHForwardTracker  classes, and maybe even carve out a dedicated section in api.rst for discussing this tracker functionality.

AlexMKX:
Thanks for your detailed response @ronf . 
Sounds great — happy to proceed with this design. tracker_factory + per-connection tracker with the SSHForwardTracker / SSHPortForwardTracker / SSHPathForwardTracker hierarchy reads cleanly and gives a clear migration path for the remote-forward and forward_socks cases later.

A few things I'd like to confirm before I rework the PR:

1. forward_local_bytes / forward_remote_bytes semantics: is this purely an observer (returns None, asyncssh ignores the return value), or a transform (return value replaces what gets forwarded)? Observer fits my idle-tracking use case; transform would be powerful but has very different implications. I'd lean observer for the initial API.

2. Are the byte hooks meant to be always called when present, or optional/opt-in? For low-overhead aggregate trackers (just counting connections), routing every byte through a Python method has a real throughput cost. A way to declare "this tracker doesn't want byte hooks" — either by leaving the method as the parent's no-op default (detected via `type(tracker).forward_local_bytes is SSHForwardTracker.forward_local_bytes`), or via an explicit attribute — would let metrics-only trackers stay cheap. Happy to follow whatever convention you prefer.

3. connection_lost(exc) vs split local_connection_lost / remote_connection_lost: I'd lean toward keeping the single hook for now, since splitting is forward-compatible (the per-direction variants can be added later without breaking existing code, as long as the single hook fires on first-side-closed). Unless you'd prefer the split to land as part of the initial API, I'll go with one method.

For scope, I'll start with local forwarding (forward_local_port + forward_local_port_to_path sharing SSHPortForwardTracker; forward_local_path + forward_local_path_to_port sharing SSHPathForwardTracker) and leave the remote-forward wiring + forward_socks for follow-up PRs, designed so they slot in without major rework.

Good call on changes.rst — I'll drop it from the PR. For the docs side, I can add docstrings to SSHForwardTracker / SSHPortForwardTracker / SSHPathForwardTracker covering each method's contract (when it's called, what the args are, return-value semantics), plus a dedicated section in api.rst that explains the factory pattern, shows a minimal example, and notes which forward_* methods accept tracker_factory. If you have a preferred section name or placement (e.g., right after the existing forwarding methods, or its own top-level section), let me know — otherwise I'll pick a reasonable spot and you can move it during review.

ronf:
> forward_local_bytes / forward_remote_bytes semantics: is this purely an observer (returns None, asyncssh ignores the return value), or a transform (return value replaces what gets forwarded)? Observer fits my idle-tracking use case; transform would be powerful but has very different implications. I'd lean observer for the initial API.

I was thinking pure observer here. Once you start thinking about transformations or parsing information out of the streams, you're probably better off calling `create_connection()` on the `SSHConnection` with your own `SSHTCPSessionFactory`, or calling `open_connection()` to get back `SSHReader` and `SSHWriter` instance you can use to manipulate the stream. Admittedly, you this would require also manually setting up your own listener to accept new connections, but that can be done with only a few lines of code. The only messy part is trying to get a common listening port if your listen host resolves to multiple IPs.

> Are the byte hooks meant to be always called when present, or optional/opt-in? For low-overhead aggregate trackers (just counting connections), routing every byte through a Python method has a real throughput cost. A way to declare "this tracker doesn't want byte hooks" — either by leaving the method as the parent's no-op default (detected via type(tracker).forward_local_bytes is SSHForwardTracker.forward_local_bytes), or via an explicit attribute — would let metrics-only trackers stay cheap. Happy to follow whatever convention you prefer.

Anything you did to query if a method was defined might not be much cheaper than making an unconditional call to an empty function body (which could be provided in the tracker parent class as you said). If we did find that the method lookup was expensive enough to matter, we could always have the tracker parent constructor scan for which methods exist and cache the result in a member variable. However, I'd want to measure the performance before attempting any optimizations of that sort. Keep in mind the byte callbacks are once per SSH message, not per byte, and applications transferring large amounts of data are probably using larger block sizes on writes. Also, all SSH messages have to go through a decryption step, which is likely to be FAR more expensive than a Python method call.

> connection_lost(exc) vs split local_connection_lost / remote_connection_lost: I'd lean toward keeping the single hook for now, since splitting is forward-compatible (the per-direction variants can be added later without breaking existing code, as long as the single hook fires on first-side-closed). Unless you'd prefer the split to land as part of the initial API, I'll go with one method.

I'm torn on this one. If we were to split it, it seems a bit asymmetric to not also do the split on `connection_made()`, but we don't want to do that because we want one tracker object for both sides of the forwarding. Also, this would end up triggering two callbacks back to back in pretty much all cases, when the only value we're getting is information on which side closed first. I'm ok going without the split for now.

> For scope, I'll start with local forwarding (forward_local_port + forward_local_port_to_path sharing SSHPortForwardTracker; forward_local_path + forward_local_path_to_port sharing SSHPathForwardTracker) and leave the remote-forward wiring + forward_socks for follow-up PRs, designed so they slot in without major rework.

Sounds good. I probably won't release the feature until at least local and remote are both done, but I like the idea of fully reviewing the local case first to make sure everything looks good. It keeps the size down and there'll be less throw-away work if changes are required.

As for `forward_socks()`, we'll see how cleanly this fits in, and we can make a call about whether to include that or not.

AlexMKX:
All three confirmations make sense and I'll go with them as the basis for the rework:

1. Pure observer for byte hooks — clean boundary. Anyone wanting transform/parsing semantics drops down to create_connection() / open_connection().

2. No opt-in machinery for now; rely on no-op defaults in the parent class. The per-message granularity plus the cost of SSH decryption makes the Python method call effectively free in any realistic workload. Revisit only if benchmarks ever say otherwise.

3. Single connection_lost — agreed. The symmetry argument with single connection_made is the right way to look at it; splitting one side without the other would feel arbitrary, and the value of knowing which side closed first is too narrow to justify the API surface.

Scope: I'll send the rework as a local-only PR (forward_local_port + forward_local_port_to_path sharing SSHPortForwardTracker; forward_local_path + forward_local_path_to_port sharing SSHPathForwardTracker), structured so adding the remote-forward methods later is a small follow-up rather than a redesign. forward_socks I'll leave alone for now and we'll decide once the base case lands.

I'll start on it now. A few small implementation-detail questions are likely to come up — whether SSHForwardTracker is an abstract base or a concrete class with no-op defaults users can subclass selectively, what happens if tracker_factory() raises, doc placement in api.rst — but those will be more productive to raise on the actual diff than upfront. I'll ping when the PR is updated.

ronf:
Generally speaking, all of the AsyncSSH factory classes are meant to be subclassed, with a default implementation of pretty much all of the methods so that only methods an application wants to change need to defined in the subclass.

In terms of placement in api.rst, I was thinking that there was already a top-level section for forwarding, but I see now that's not the case. Instead, there's a paragraph in the "Overview" section about forwarding methods, and then a "Forwarder Classes" section. The new tracker classes can probably be added in "Forwarder Classes" below SSHForwarder, with the details of each callback being described in the method docs, similar to what is done in SSHClient and SSHServer. It's probably also worth a quick mention of these new classes in the Overview section just after the existing forwarding summary.

All of the methods can be listed for the Port and Path versions of the tracker classes, but the implementation of those will mostly be coming from the definition in the parent class in most cases. The exception would be connection_made(), as its arguments will be different. This is similar to SSHTCPSession and SSHUNIXSession.

AlexMKX:
Reworked the PR to match what we landed on above and force-pushed `feat/forward-tracker`:

- `tracker_factory` (per-connection) instead of a single shared tracker, with the `SSHForwardTracker` / `SSHPortForwardTracker` / `SSHPathForwardTracker` hierarchy mirroring the forwarder classes. `connection_made(forwarder, orig_host, orig_port)` for TCP, `connection_made(forwarder)` for UNIX.
- Observer-only `forward_local_bytes` / `forward_remote_bytes` (no-op defaults on the base, no opt-in machinery), single `connection_lost(exc)`.
- Wired into all four local methods (`forward_local_port`, `forward_local_port_to_path`, `forward_local_path`, `forward_local_path_to_port`). Remote forwards and `forward_socks` deliberately left out for a follow-up.
- Dropped `changes.rst`; added a "Forward Tracker Classes" section to `api.rst`.

One implementation detail worth flagging for review: `connection_lost` clears the per-connection tracker on first call so it fires exactly once — `_forward()` notifies manually on a `ChannelOpenError`, and the local transport's subsequent close would otherwise deliver a second `connection_lost`. There's a regression test for that ("lost fires exactly once" on an accept_handler denial).

The PR description above is updated to the new design. Happy to adjust naming, hook signatures, or the api.rst placement on the diff.

asyncssh/__init__.py
```diff
@@ -40,7 +40,8 @@
 
 from .config import ConfigParseError
 
-from .forward import SSHForwarder
+from .forward import SSHForwarder, SSHForwardTracker
```
ronf: I'm not sure we need to expose SSHForwardTracker as a public symbol. Callers should always base their trackers on either the Port or Path version, depending on whether the non-tunneled traffic is over TCP or UNIX domain sockets.

asyncssh/__init__.py
```diff
@@ -151,7 +148,8 @@
     'SSHAgentKeyPair', 'SSHAuthorizedKeys', 'SSHCertificate', 'SSHClient',
     'SSHClientChannel', 'SSHClientConnection', 'SSHClientConnectionOptions',
     'SSHClientProcess', 'SSHClientSession', 'SSHCompletedProcess',
-    'SSHForwarder', 'SSHKey', 'SSHKeyPair', 'SSHKnownHosts',
+    'SSHForwarder', 'SSHForwardTracker', 'SSHPortForwardTracker',
```
ronf: See above.

asyncssh/connection.py
```diff
@@ -3210,7 +3210,9 @@ async def forward_unix_connection(self, dest_path: str) -> SSHForwarder:
     async def forward_local_port(
             self, listen_host: str, listen_port: int,
             dest_host: str, dest_port: int,
-            accept_handler: Optional[SSHAcceptHandler] = None) -> SSHListener:
+            accept_handler: Optional[SSHAcceptHandler] = None,
+            tracker_factory:
+                Optional[SSHForwardTrackerFactory] = None) -> SSHListener:
```
ronf: I think we'll probably need factory types SSHPortForwardTrackerFactory and SSHPathForwardTrackerFactory instead of a single SSHForwardTrackerFactory. Which one you use will be determined by whether the listener is TCP or UNIX, which we'll generally know by what function is being called to set up the forwarding. For instance, forward_local_port() would need the Port version of the tracker factory.

asyncssh/connection.py
```diff
@@ -3233,11 +3235,18 @@ async def forward_local_port(
...
+               which returns a new :class:`SSHPortForwardTracker` (or
+               :class:`SSHForwardTracker` subclass) for observing that
+               connection's lifecycle. `None` (default) disables tracking
+               with no overhead.
            :type listen_host: `str`
            :type listen_port: `int`
            :type dest_host: `str`
            :type dest_port: `int`
            :type accept_handler: `callable` or coroutine
+           :type tracker_factory: `callable` or `None`
```
ronf: For arguments which are optional because they default to `None`, I generally don't call out `None` in the "type" description. The argument would just be left out if a caller didn't want tracker, so there's no reason for them to pass in `None` explicitly.

We could also consider using the factory type name as the value under "type", since we have one defined.

asyncssh/connection.py
```diff
@@ -3278,10 +3287,9 @@ async def tunnel_connection(
                              (dest_host, dest_port))
 
         try:
-            listener = await create_tcp_forward_listener(self, self._loop,
-                                                         tunnel_connection,
-                                                         listen_host,
-                                                         listen_port)
+            listener = await create_tcp_forward_listener(
+                self, self._loop, tunnel_connection, listen_host, listen_port,
+                tracker_factory=tracker_factory)
```
ronf: There's no need to make this a keyword argument. It should be fine to keep it positional.

asyncssh/connection.py
```diff
@@ -3233,11 +3235,18 @@ async def forward_local_port(
                or not to allow connection forwarding, returning `True` to
                accept the connection and begin forwarding or `False` to
                reject and close it.
+           :param tracker_factory:
+               An optional callable invoked once per accepted connection
+               which returns a new :class:`SSHPortForwardTracker` (or
+               :class:`SSHForwardTracker` subclass) for observing that
```
ronf: I don't think you need the "or" here on SSHForwardTracker. The tracker_factory should be an SSHPortForwardTracker, or a subclass of that.

asyncssh/connection.py
```diff
@@ -3297,8 +3305,10 @@ async def tunnel_connection(
         return listener
 
     @async_context_manager
-    async def forward_local_path(self, listen_path: str,
-                                 dest_path: str) -> SSHListener:
+    async def forward_local_path(
```
ronf, line 3376: Same changes as above, but with the Path version of the tracker.

asyncssh/connection.py
```diff
@@ -5304,7 +5321,9 @@ async def open_tap(self, *args: object, **kwargs: object) -> \
     @async_context_manager
     async def forward_local_port_to_path(
```
ronf, line 5389: This would be the Port version.

asyncssh/connection.py
```diff
@@ -5378,9 +5403,10 @@ async def tunnel_connection(
         return listener
 
     @async_context_manager
-    async def forward_local_path_to_port(self, listen_path: str,
-                                         dest_host: str,
-                                         dest_port: int) -> SSHListener:
+    async def forward_local_path_to_port(
```
ronf, line 5472: ...and this would be the Path one.

asyncssh/forward.py
```diff
@@ -38,6 +38,113 @@
 SSHForwarderCoro = Callable[..., Awaitable]
 
 
+class SSHForwardTracker:
+    """Base class for observing the lifecycle of a forwarded connection
+
+       A tracker observes a single forwarded connection on a local
```
ronf: I would avoid mentioning "local" here, as I think we eventually want to reuse these tracker classes when creating for remote listening ports. The only place I'd expect "local" and "remote" to appear here would be for the callbacks for forwarding bytes, where we distinguish if the bytes were generated locally or remotely, independent of the direction the connection was originally set up.

asyncssh/forward.py
```diff
@@ -38,6 +38,113 @@
...
+        """Called when a new UNIX domain connection is accepted
+
+           :param forwarder:
+               The forwarder handling this connection.
+           :type forwarder: :class:`SSHForwarder`
+
+        """
+
+
+SSHForwardTrackerFactory = Callable[[], SSHForwardTracker]
```
ronf: See my comment above about having two different factory classes for Port and Path, and not exposing SSHForwardTracker or defining SSHForwardTrackerFactory.

asyncssh/forward.py
```diff
@@ -192,10 +299,69 @@ def close(self) -> None:
...
-    def __init__(self, conn: 'SSHConnection', coro: SSHForwarderCoro):
+    def __init__(self, conn: 'SSHConnection', coro: SSHForwarderCoro,
+                 tracker_factory: Optional[SSHForwardTrackerFactory] = None):
         super().__init__()
         self._conn = conn
         self._coro = coro
+        self._tracker_factory = tracker_factory
+        self._tracker: Optional[SSHForwardTracker] = None
+
+    def _create_tracker(self) -> None:
```
ronf: You might want to think about having _create_tracker() take the tracker_factory as an argument, rather than assigning that as a member of the class, as we shouldn't need it after the tracker is created.

I'm also thinking we can make the type of the self._tracker member be a generic type tied to the type of the tracker factory passed in, so you don't need to downcast it in the child classes.

It's probably also worth creating a utility function which does the try/except on Exception, so that doesn't need to be repeated in so many places.

asyncssh/forward.py
```diff
@@ -234,11 +400,21 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
 
         super().connection_made(transport)
 
+        orig_host, orig_port = '', 0
```
ronf: I would suggest doing this as an "else" clause on the "if" below. In practice, this doesn't seem to ever happen, so it probably will need to be marked with a "# pragma: no cover" just like the "if" currently has a "# pragma: no branch".

docs/api.rst
```diff
@@ -1009,6 +1009,70 @@ Forwarder Classes
    ============================== =
 
 
+Forward Tracker Classes
+=======================
+
+The ``forward_local_*`` methods on :class:`SSHClientConnection` accept an
```
ronf: This will need to be updated once we add tracker support for the forward_remote_* methods.

docs/api.rst
```diff
@@ -1009,6 +1009,70 @@ Forwarder Classes
...
+              self._counter.active += 1
+
+          def connection_lost(self, exc):
+              self._counter.active -= 1
+
+      listener = await conn.forward_local_port(
+          '', 0, 'remote-host', 80,
+          tracker_factory=lambda: ConnCounter(counter))
+
+.. autoclass:: SSHForwardTracker()
```
ronf: I think we can get away with listing only the Port and Path classes here, and just repeating the parent class methods under each of the child classes here. You can see other examples of that in the various flavors of SSHSession, for instance. Note that SSHSession itself is not listed in api.rst.

tests/test_forward.py
```diff
@@ -697,6 +697,162 @@ async def accept_handler(_orig_host: str, _orig_port: int) -> bool:
                 writer.close()
                 await maybe_wait_closed(writer)
 
+    @asynctest
+    async def test_forward_local_port_tracker_factory_fires_made_and_lost(self):
```
ronf: This name is a bit on the long side. 😄

tests/test_forward.py
```diff
@@ -697,6 +697,162 @@ async def accept_handler(_orig_host: str, _orig_port: int) -> bool:
...
+        self.assertIn('lost', kinds)
+
+        made = next(event for event in events if event[0] == 'made')
+        self.assertIsInstance(made[1], asyncssh.SSHForwarder)
+        self.assertEqual(made[2], '127.0.0.1')
+        self.assertIsInstance(made[3], int)
+
+    @asynctest
+    async def test_forward_local_port_tracker_factory_per_connection(self):
+        """The factory is called once per accepted connection"""
```
ronf: This should describe what this test method does, not the factory. The name of the method also needs some work.

asyncssh/forward.py
```diff
@@ -192,10 +299,69 @@ def close(self) -> None:
...
+
+    def _create_tracker(self) -> None:
+        """Instantiate this connection's tracker from the factory, if any"""
+
+        if self._tracker_factory is None:
+            return
+
+        try:
+            self._tracker = self._tracker_factory()
+        except Exception: # pylint: disable=broad-exception-caught
```
ronf: I'm running an older version of pylint here at the moment, and it isn't recognizing "broad-exception-caught". Would you mind switching this back to its original name of "broad-except"?

tests/test_forward.py
```diff
@@ -697,6 +697,162 @@ async def accept_handler(_orig_host: str, _orig_port: int) -> bool:
...
                 await maybe_wait_closed(writer)
 
+    @asynctest
+    async def test_forward_local_port_tracker_factory_fires_made_and_lost(self):
+        """A port tracker sees connection_made and connection_lost"""
+
+        events = []
+        lost = asyncio.Event()
+
+        class _RecordingTracker(asyncssh.SSHPortForwardTracker):
```
ronf, line 748: These tracker classes are missing docstrings, triggering a pylint error (6 places total).

tests/test_forward.py
```diff
@@ -697,6 +697,162 @@ async def accept_handler(_orig_host: str, _orig_port: int) -> bool:
...
+           a manual notify in _forward() followed by the asyncio close path."""
+
+        lost_count = 0
+
+        class _Counting(asyncssh.SSHPortForwardTracker):
+            def connection_lost(self, exc):
+                nonlocal lost_count
+                lost_count += 1
+
+        async def deny(_h, _p):
```
ronf: The short variable names here are triggering pylon's invalid-name error.

asyncssh/forward.py
```diff
@@ -249,4 +425,13 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
         """Handle a newly opened connection"""
 
         super().connection_made(transport)
+
+        self._create_tracker()
+
+        if self._tracker is not None:
+            try:
+                cast(SSHPathForwardTracker, self._tracker).connection_made(self)
+            except Exception: # pylint: disable=broad-exception-caught
```
ronf: The test code doesn't seem to be covering the case of connection_made triggering an Exception. I think that is only being checked for the Port case right now and not the Path case, which has a different connection_made() method.

ronf (changes requested):
This looks great! I've added inline comments, but they're all for pretty minor issues.

asyncssh/__init__.py
ronf: Unlike other imports where I sometime choose logical ordering over pure alphabetical, the "__all__" definition is one place where I generally go for pure alphabetical (case-sensitive), putting Path before Port on the tracker classes. Also, there's room to insert SSHForwarder, SSHPathForwardTracker, and SSHPortForwardTracker all on the same line.

asyncssh/connection.py
ronf: Looks like this comment still applies to the update.

asyncssh/forward.py
ronf: Coverage now looks good.

ronf:
Just reviewed your update -- looks good! Based on the changes, I marked most of my previous comments as resolved. There's only a handful left.

asyncssh/forward.py
AlexMKX: Done — reworded the docstring to drop the "local listener" framing so it stays accurate for future remote-port tracker reuse.

asyncssh/forward.py
AlexMKX: Done — `_create_tracker()` now takes the factory as an argument (called once from `__init__`, no longer stored as a member), and `SSHLocalForwarder` is generic over the tracker type so `self._tracker` no longer needs downcasting in `SSHLocalPortForwarder`/`SSHLocalPathForwarder`.

asyncssh/forward.py
AlexMKX: Done — moved the '', 0 default into an `else` clause with `# pragma: no cover`, mirroring the `if`'s `# pragma: no branch`.

tests/test_forward.py
AlexMKX: Shortened to `test_port_tracker_made_and_lost`.

docs/api.rst
AlexMKX: Acknowledged — the api.rst forwarding section will be updated when tracker support lands for the forward_remote_* methods; leaving this open as a reminder.

AlexMKX:
> Just reviewed your update -- looks good! Based on the changes, I marked most of my previous comments as resolved. There's only a handful left.

It should be better now. Are there anything to fix?

asyncssh/forward.py
AlexMKX: Also done — the repeated try/except around tracker hooks is now a single `_notify_tracker` helper.

asyncssh/forward.py
```diff
@@ -297,52 +299,55 @@ def close(self) -> None:
             peer.close()
 
 
-class SSHLocalForwarder(SSHForwarder):
+class SSHLocalForwarder(SSHForwarder, Generic[_Tracker]):
```
ronf: What's driving the requirement for Generic here? Why do you need the type for _tracker to be the subclass rather than SSHForwardTracker? Is this related to the new notify helper?

asyncssh/forward.py
```diff
@@ -297,52 +299,55 @@ def close(self) -> None:
...
+                notify(tracker)
             except Exception: # pylint: disable=broad-except
                 pass
 
+    def data_received(self, data: bytes,
+                      datatype: Optional[int] = None) -> None:
+        """Handle incoming data from the local transport"""
+
+        self._notify_tracker(
+            self._tracker, lambda t: t.forward_local_bytes(data))
```
ronf: Rather than creating a lambda function here, I wonder if you could pass in one of the bound tracker methods rather than the tracker object itself as the first argument. You'd probably get less type safety, passing the arguments through an "args" argument, but it would avoid the need for the lambda.

An alternative would be to use a nested function definition for the wrapper instead of a lambda, like what's done in forward.py today for defining session_factory in the _forward() method. I think that's a bit easier to read.

tests/test_forward.py
```diff
@@ -698,7 +698,7 @@ async def accept_handler(_orig_host: str, _orig_port: int) -> bool:
                 await maybe_wait_closed(writer)
 
     @asynctest
-    async def test_forward_local_port_tracker_factory_fires_made_and_lost(self):
+    async def test_port_tracker_made_and_lost(self):
```
ronf, line 742: Looks good. Some of the other function names here could probably also use some trimming.

ronf (commented):
A few comments on the latest changes.

asyncssh/forward.py
ronf: Also, pylint is raising a no-self-use error on _notify_tracker(). One option would be to make that a @staticmethod.

asyncssh/forward.py
ronf: Ah - I think I see. This avoids the downcasting for the connection_made() tracker calls. Depending on what we do with _notify_tracker, this may not be an issue. Even now, I wonder if you're actually getting any type checking on these calls, given the lambda wrapper without a specific type declared for t.

asyncssh/forward.py
AlexMKX: Switched to nested function definitions instead of lambdas, matching the `session_factory` style used in `_forward()`. Each `notify()` is annotated with the concrete tracker type at that call site (`SSHPortForwardTracker` / `SSHPathForwardTracker` for the `connection_made` hooks, `_Tracker` for the shared byte/`connection_lost` hooks in the base class).

AlexMKX: `_notify_tracker` is now a `@staticmethod` (generic over `_Tracker`, still called as `self._notify_tracker(...)`). Confirmed with pylint (using the repo `pylintrc` + the `no_self_use` extension, which is what surfaces this on newer pylint) that the `no-self-use` finding is gone.

asyncssh/forward.py
AlexMKX: Right — `Generic[_Tracker]` is what lets `self._tracker` carry the concrete subclass type in each child forwarder, per your earlier suggestion (in #3487294241) to avoid the downcasts. Removing it would bring the `cast()`s back.

On the type-checking doubt: with the nested `notify()` defs, each hook parameter now carries a concrete tracker type, so mypy does check these calls against `self._tracker`. E.g. annotating `SSHLocalPortForwarder`'s `notify()` with `SSHPathForwardTracker` by mistake gives:

`error: Argument 2 to "_notify_tracker" of "SSHLocalForwarder" has incompatible type "Callable[[SSHPathForwardTracker], None]"; expected "Callable[[SSHPortForwardTracker], None]"  [arg-type]`

AlexMKX: One correction for the record: I went back and checked the old lambda form too, and it was actually type-checked as well — mypy infers the lambda parameter from the `Callable[[_Tracker], None]` signature once `_Tracker` is bound by the `tracker` argument (`reveal_type` resolves it to the concrete subclass), and the same deliberate wrong-arity call was flagged there too. So the type safety comes from `Generic[_Tracker]` in both forms; the move to nested defs is purely the readability win you suggested, not a correctness fix.

ronf: Good to know - thanks for looking into that...

tests/test_forward.py
ronf: These changes all look good and I'm seeing clean runs on mypy, pylint, and tox (including coverage).

If you're happy with this code I think the next step is to look into adding the "remote" versions of this tracker support. I'm in the middle of addressing some bug reports in the "develop" branch right now, but hopefully by the time the remote support is ready to go, things will quiet down and I'll be able to get this into the next feature release...

Thanks for all your work on this so far!

AlexMKX: Thanks! Yes, I am happy with the current state, so I will start on the remote versions.

One design question before I get too far, about the tracker lifecycle boundary: for a connection accepted by the remote listener whose local destination socket cannot be opened, should `tracker_factory` still be called, and should that tracker then get `connection_made` followed by `connection_lost(...)`? I lean towards yes, so the lifecycle mirrors local forwarding (where the tracker is created for the accepted socket, before the SSH channel open can fail) and there is exactly one tracker per accepted listener connection. Happy to go the other way if you would rather a tracker only appear once both ends are established.

AlexMKX:
Pushed the remote half. Commits `bc19857..96624f4` add `tracker_factory` to the four `forward_remote_*` methods, with the tracker plumbing lifted into a shared base so the remote forwarders reuse it rather than duplicating it. I have updated the PR title and description to match — the old text said remote forwarding was out of scope, which is no longer true.

To be explicit about review state: your 1 Aug pass applies up to `2bf4164`. Everything after that is new and unreviewed.

Two things worth flagging:

- `pylintrc` `max-module-lines` 10000 -> 11000, because `connection.py` was 26 lines under the cap and the four new signatures do not fit under it. Easy to revert if you would rather take the C0302.
- The tracker lifecycle question for a destination that cannot be opened is in the thread above. Current behaviour: the tracker is created when the listener connection is accepted, so it gets `connection_made` and then `connection_lost(ChannelOpenError)`. Flipping it is a two-line change in each of the two helpers.

Gates on my side: 192 passed in `test_forward.py`, pylint 9.81 (unchanged message set), mypy clean on `forward.py`, sphinx builds warning-free.

tests/test_forward.py
ronf: Yeah, I think this makes sense. We'd want to know about the incoming connections even if the forwarding target isn't successful. Even if we deferred `connection_made()` until after finishing the outbound connection, there's no way to report the exception in `connection_made()`, so it's probably simpler to wait and report that exception in `connection_lost()`. That also means the generated callbacks look the same regardless of whether the failure is detected before or after the call to `connection_made()`.

ronf:
Sorry for the slow response... I'm still trying to address some other feedback and may not get back to this for a few weeks. I did actually start going through the diff here, but it was a bigger set of changes than I was expecting. I know a lot of it is just new test cases, which is fine, but I wasn't expecting things like the new _forward_tcp_connection and _forward_unix_connection functions. I haven't really had a chance to go through them carefully, though, to see what changed from the original forward_connection() and forward_unix_connection(). The intent with those functions was that they could be called by the application developer, if the caller wanted to do some validation or modification of where to forward to, or the library would call those functions for the caller if they just returned `True`. 

I also haven't quite figured out why SSHRemotePortForwarder and SSHRemotePathForwarder were needed. I'm not sure anything like those exist in the local port forwarding case.

AlexMKX:
Thanks for the frank feedback, and no worries on the timing. You were right on both points, so I've reworked the remote half rather than defend it.

I over-read your lifecycle answer as requiring `connection_made()` before the outbound connect. That led to the pre-built forwarder, the two private helpers, and the remote forwarder classes on the SSH-channel side, which in turn needed inverted byte hooks. None of that was necessary.

The new version (`6b77ab8..13d125a`, replacing `bc19857..96624f4`) puts the tracker on the forwarder for the local destination socket, the same way local forwarding does:

- The byte hooks map exactly as they do there, and there are no new forwarder classes.
- `connection_made()` is called once the destination is connected, so the forwarder it's given is fully usable, including `close()`.
- If the destination can't be opened, the tracker gets `connection_made()` followed by `connection_lost(exc)`, as we discussed, and the `ChannelOpenError` propagates as before.
- `forward_connection()` and `forward_unix_connection()` are unchanged apart from one optional trailing argument, `peer_factory`, which supplies the tracked protocol. With no tracker, the remote path is the original code.
- The previously approved local code is untouched apart from docstrings.
- The production diff is now +160/−42.

Two things to flag:

- `connection.py` on `develop` is already at exactly 10000 lines, the `max-module-lines` limit, so any addition trips C0302. This branch is at 10014. I've left `pylintrc` alone; it's your call whether to raise the limit or handle it another way.
- Pre-existing, not new here: if the SSH connection closes while a remote-forward destination connect is still pending, and that connect later succeeds, the destination socket is left open (`_finish_open_request()` drops the session). With a tracker, this shows up as `connection_made()` with no `connection_lost()`. Happy to look at that separately, like #823.

Issue #811, opened 2026-06-11
If Included configuration ends with Host block, further configurations are not processed


tazle:
I have an SSH configuration composed of multiple files, all included in the regular `~/.ssh/config` like so:

```
Include ~/.colima/ssh_config

Include ~/.ssh/config_work
Include ~/.ssh/config_home

Include ~/a/project/some_auto_generated_config

```

OpenSSH processes each of the includes, but asyncssh only the first one.

Debug logs (with some custom debugging added, the "Reading" entries are standard):
```
DEBUG:asyncssh:Reading config from "/Users/tuure/.ssh/config"
DEBUG:asyncssh:Handling line "Include /Users/tuure/.colima/ssh_config"
DEBUG:asyncssh:Including "/Users/tuure/.colima/ssh_config"
DEBUG:asyncssh:Reading config from "/Users/tuure/.colima/ssh_config"
DEBUG:asyncssh:Handling line "Host colima"
DEBUG:asyncssh:Continuing reading "/Users/tuure/.ssh/config"
```

and this is where it ends, `~/.ssh/config_work` and friends never get processed.

This is because the configuration files typically end with a `Host` block, and `self._matching` is not restored properly after a Host block if the file ends. `_include` needs to restore `self._matching` in addition to `self._path`. Also probably `self._line_no` would be good to restore.

ronf:
I think this is a duplicate of #810. Could you try the fix I suggested there and see if it works for you? It would look something like:

```patch
diff --git a/asyncssh/config.py b/asyncssh/config.py
index d67e4e9..764d1b7 100644
--- a/asyncssh/config.py
+++ b/asyncssh/config.py
@@ -173,6 +173,7 @@ class SSHConfig:
             for path in paths:
                 self.parse(path)

+        self._matching = True
         self._path = old_path
         args.clear()

```

tazle:
I saved `self._matching` into `old_matching` and restored it like `old_path` (also did the same for line number), and that seemed to work.

ronf:
Good catch on restoring line_no along with the path.

Regarding matching, you don't really need to save "old_matching" here, as you'd only ever get into the `_include()` function when `self._matching` was set to `True`. So, while it needs to be restored in case there's content left to parse in the parent file after the Include directive, it should be sufficient to just set `self._matching` back to `True`.

In the case where there was an Include inside a Host block that was skipped, you'd never enter the `_include()` function.

ronf:
[...]

PR #812, opened 2026-06-12
Save and restore matching state and line number across include

tazle:
This was my fix to #811. ~I didn't test this particular version yet, though.~ I did now, on top of 2.22.0, seemed to work.


ronf:
Thanks - I went with a slight variation of this change, and also updated a unit test to cover this.

PR #821, opened 2026-07-10
Auto-reformat with `ruff check --select=I --fix`

waketzheng:
# What does this PR do

Add `ruff` to pre-commit config and run `prek run --all-files`, which produces the same result as `ruff check --select=I --fix`.

# Why this PR

To improve the contributor experience.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,3 +18,10 @@ repos:
         additional_dependencies:
           - tomli
         exclude: '^docs/changes\.rst$|^\.?pylintrc$'
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: 'v0.15.21'
+    hooks:
+      - id: ruff
+        args: [--select=I, --fix, --exit-non-zero-on-fix, --show-fixes]
+        types: [python]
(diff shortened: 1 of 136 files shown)


ronf:
Sorry, this is too big a diff for me to accept. Also, I don't really agree with a lot of the reformatting that ruff has chosen to do here. If there are specific findings which are not pure formatting, I'd be willing to consider them, but even there it would require some time to validate each change, so it needs to be a more manageable size.

waketzheng:
Thanks for the honest feedback. This project is useful and easy to use.

How do you check type hints? I noticed that running `mypy asyncssh` outputs a lot of errors.

ronf:
I'm getting a clean run against mypy right now, with mypy version 1.17.1 using the following command:

```
env MYPYPATH=stubs mypy --html-report /tmp/cov asyncssh
```

This version of mypy is about a year old at this point. Every now and then I update to the latest mypy and usually have to make a small number of fixes in the asyncssh code to adjust for the new version.

I do something similar with pylint, but it's been a while since I upgraded that. At the moment, I'm on pylint 2.9.3, which is like 5 years old. I really should refresh that at some point.

PR #825, opened 2026-08-04
Apply `X | Y` type hints style for asyncssh/sk.py

waketzheng:
# Description

The code was automatically changed by running `ruff check --select=UP --fix asyncssh/sk.py` with Ruff version 0.16.1.

### DIFF
--- a/asyncssh/sk.py
+++ b/asyncssh/sk.py
@@ -25,12 +25,12 @@ import ctypes
 from hashlib import sha256
 import hmac
 import time
-from typing import Callable, List, Mapping, NoReturn, Optional
-from typing import Sequence, Tuple, TypeVar, cast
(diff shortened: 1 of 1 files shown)


ronf:
Now that the minimum supported version of Python for AsyncSSH has moved to 3.10, migrating to this new syntax is an option, and definitely something I'd consider.

That said, I appreciate your effort, but instead of opening a PR I would suggest opening a discussion thread with suggestions for what tool and arguments can be used to do the automated conversion (like you included in this PR). I'd have to do a line-by-line validation anyway that the tool didn't affect anything other than typing information, and I'd have more confidence doing that when running such a tool first-hand.

This applies to any automated formatting or syntax upgrading.

